PR #42, opened 2022-02-07
Regenerate with jhipster 7.2.0 and react-mobx blueprint

calvinlu3:
This is related to https://github.com/oncokb/oncokb/issues/2944
- Downgraded to jhipster 7.2, so we can use [generator-jhipster-react-mobx](https://github.com/erkieh/generator-jhipster-react-mobx)
- Generated with skipUserManagement and skipFakeData set to false
- [x] Delete some unnecessary files (ie registration, etc)


calvinlu3:
Closing because found a way to generate with jhipster 7.3.1

calvinlu3:
@zhx828 When you have the chance could you take a look at the generated files? 

- The issue with `history` that I brought up during standup has to do with the compatibility of `react-router-dom@5 and history@5`.

pom.xml
```diff
@@ -30,9 +30,8 @@
         <!-- Build properties -->
         <maven.version>3.3.9</maven.version>
         <java.version>11</java.version>
-        <!-- we use nvm manages the npm version. npm 6 is the default for node 14 -->
         <node.version>v14.17.6</node.version>
-        <npm.version>6.14.15</npm.version>
+        <npm.version>7.23.0</npm.version>
```
zhx828, line 34: I remember there was an issue using v7 with node 14. Seems no longer the issue?

calvinlu3: Thanks for pointing that out, I think the npm version change was a result of jhipster generator. Let me double check

PR #44, opened 2022-02-16
Create initial routes and navigation sidebar

calvinlu3:
This is related to https://github.com/oncokb/oncokb/issues/2942 and https://github.com/oncokb/oncokb/issues/2945


package.json
```diff
@@ -93,6 +96,8 @@
     "react-redux": "7.2.5",
     "react-redux-loading-bar": "5.0.2",
```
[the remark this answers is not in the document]
calvinlu3: The frontend tests were using redux, but those might just be leftovers from the old transcript app @7.3.1. I can take a look at the tests and remove if they don't apply anymore.

src/main/webapp/app/components/sidebar/SideBar.tsx
```diff
@@ -0,0 +1,65 @@
...
+    return (
+      <NavLink to={props.linkTo} className="nav-item" exact>
+        {menuItem}
+      </NavLink>
+    );
+  }
+  return menuItem;
+};
+
+class SideBar extends React.Component<StoreProps> {
```
zhx828, line 38: We can start with the simple implementation, but later when we start to implement the submenu, could consider https://github.com/azouaoui-med/react-pro-sidebar or other options.

src/main/webapp/app/oncokb-commons/styles/variables.scss
```diff
@@ -0,0 +1,35 @@
...
+/* Overwrite the bootstrap defaults */
+// $font-family-sans-serif: 'Gotham Book', serif;
+// color overwrite
+$blue: $oncokb-blue;
+$grey: #808080; // grey
+$light-grey: #d3d3d3; // light grey
+$primary: $oncokb-blue;
+$secondary: $grey;
+$warning: #ffc107;
+$danger: #dc3545;
```
zhx828, line 14: Worth moving these to oncokb-styles. Don't have to do it in this pr though.

zhx828 (approved):
Looks good。

src/main/webapp/app/components/sidebar/SideBar.tsx
calvinlu3: Another option is https://www.devwares.com/docs/contrast/react/navigation/sidebar/. I will try both out.

PR #47, opened 2022-03-08
Account login

calvinlu3:
This is related to https://github.com/oncokb/oncokb/issues/2943

- Can use https://github.com/calvinlu3/oncokb-keycloak to setup keycloak server with mysql and custom login theme
- Setup Google SSO
- Create keycloak custom theme that removes username/password form
- Store spring session in redis


src/main/java/org/mskcc/oncokb/transcript/service/UserService.java
```diff
@@ -78,10 +87,125 @@ public void deleteUser(String login) {
...
+
+    private static User getUser(Map<String, Object> details) {
+        User user = new User();
+        Boolean activated = Boolean.TRUE;
+        // handle resource server JWT, where sub claim is email and uid is ID
+        if (details.get("uid") != null) {
+            user.setId((String) details.get("uid"));
+            user.setLogin((String) details.get("sub"));
+        } else {
+            user.setId((String) details.get("sub"));
```
zhx828, line 165: I think sub should always be assigned to login. More to discuss.

src/main/java/org/mskcc/oncokb/transcript/service/UserService.java
```diff
@@ -78,10 +87,125 @@ public void deleteUser(String login) {
...
+     * Returns the user from an OAuth 2.0 login or resource server with JWT.
+     * Synchronizes the user in the local repository.
+     *
+     * @param authToken the authentication token.
+     * @return the user from the authentication.
+     */
+    @Transactional
+    public AdminUserDTO getUserFromAuthentication(AbstractAuthenticationToken authToken) {
+        Map<String, Object> attributes;
+        if (authToken instanceof OAuth2AuthenticationToken) {
```
zhx828, line 132: How does the JWT Auth work now? I tried to use a self generated token, but I think the class is UsernamePasswordAuthenticationToken.

zhx828 (changes requested):
- The user id field changed to String which is a big change. I see the oauth will assign the session id to the id field. But is it necessary?
- Let's do a global search and replace for the keycloak port?

calvinlu3:
- Removed jwtDecoder and CustomClaimsConverter (and other files) that are not used
- Only users that are in application db and activated will be allowed to use the site
- Fixed issue with custom JWT not working with the GET /api/account endpoint
- Changed User ID field back to Long type
- Change keycloak port to 8080 and adjusted jhipster naming

src/main/java/org/mskcc/oncokb/transcript/service/dto/UserDTO.java
```diff
@@ -1,24 +1,71 @@
...
     private String login;
 
+    @Size(max = 50)
+    private String firstName;
+
+    @Size(max = 50)
+    private String lastName;
+
+    @Email
+    @Size(min = 5, max = 254)
```
zhx828, line 30: We should double check the Size limits vs liquibase definition. This email here allows 254, but liquibase has it as 191. Could you go over all limitations and see what causes the discrepancy ?

zhx828 (approved):
Once you fixed all my comments, feel free to merge.

src/main/java/org/mskcc/oncokb/transcript/security/CustomOAuthSuccessHandler.java
```diff
@@ -0,0 +1,50 @@
...
+    public void onAuthenticationSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication)
+        throws IOException, ServletException {
+        Map<String, Object> attributes = ((OAuth2AuthenticationToken) authentication).getPrincipal().getAttributes();
+        Gson gson = new Gson();
+        String json = gson.toJson(attributes);
+        KeycloakUserDTO keycloakUser = gson.fromJson(json, KeycloakUserDTO.class);
+
+        Optional<User> user = userRepository.findOneByEmailIgnoreCase(keycloakUser.getEmail());
+        // If user is in db, then let spring security handle authentication success.
+        if (user.isPresent() && user.get().isActivated()) {
```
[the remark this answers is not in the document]
calvinlu3: We were going to add a ROLE_CURATOR to public, so we can update this to check for that role when we use that table.

I will make a note of this in a comment.

src/main/java/org/mskcc/oncokb/transcript/service/dto/UserDTO.java
calvinlu3: This seems to be related to [issue from jhipster](https://github.com/jhipster/generator-jhipster/issues/8604). 
Since we're using utf8, then the max length for our unique email field can be 255 characters instead of 191. Should be okay to increase the liquibase constraint to 255.

src/main/java/org/mskcc/oncokb/transcript/service/UserService.java
```diff
@@ -78,10 +91,56 @@ public void deleteUser(String login) {
...
+     * @return the user from the authentication.
+     */
+    @Transactional
+    public Optional<UserDTO> getUserFromAuthentication(AbstractAuthenticationToken authToken) {
+        if (authToken instanceof UsernamePasswordAuthenticationToken) { // Our custom JWT
+            User user = new User();
+            user.setAuthorities(getAuthoritiesFromAuthToken(authToken));
+            user.setLogin(authToken.getName());
+            user.setActivated(true);
+            return Optional.of(new UserDTO(user));
```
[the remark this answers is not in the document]
calvinlu3: I should check if the user exists in db here.

src/main/docker/realm-config/oncokb-curation-realm.json
```diff
@@ -0,0 +1,2213 @@
...
+            "jsonType.label": "String"
+          }
+        }
+      ],
+      "defaultClientScopes": ["web-origins", "roles", "profile", "email"],
+      "optionalClientScopes": ["address", "phone", "offline_access", "microprofile-jwt"]
+    },
+    {
+      "id": "1acf7ad0-68cb-46a6-a3e4-8b2f2abecb85",
+      "clientId": "jhipster-control-center",
```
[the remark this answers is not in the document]
calvinlu3: Do we need the [jhipster control center](https://www.jhipster.tech/jhipster-control-center/)? We're not using this feature from jhipster, so can delete this client.

There are still a couple jhipster names in this file. Will see which ones are needed, if not then they can be deleted.

PR #53, opened 2022-03-29
Add Alteration/CancerType models && Update FdaSubmission Model

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/3042 and https://github.com/oncokb/oncokb/issues/3041

### DIFF
--- a/jhipster-jdl.jdl
+++ b/jhipster-jdl.jdl
@@ -71,39 +71,107 @@ entity GeneAlias {
     name String
 }
 
-entity CompanionDiagnosticDevice {
-	name String
-    manufacturer String
+entity SpecimenType{
+	type String required
 }
 
-entity PremarketApproval {
-	device String
-    genericName String
-    applicant TextBlob
-    number String
+entity CompanionDiagnosticDevice {
+	name String required
+    manufacturer String required
+}
+
+enum FdaSubmissionTypeKey {
+	PMA, DE_NOVO, HDE, PMN
+}
+
+entity FdaSubmissionType {
+	key FdaSubmissionTypeKey required
+    name String required
+    shortName String
+    descrption TextBlob
+}
+
+entity FdaSubmission {
+    number String required
     supplementNumber String
+
+	deviceName String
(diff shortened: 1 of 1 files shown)


jhipster-jdl.jdl
```diff
@@ -76,34 +76,93 @@ entity CompanionDiagnosticDevice {
     manufacturer String
 }
 
-entity PremarketApproval {
-	device String
-    genericName String
-    applicant TextBlob
+enum FdaSubmissionType {
+	PMA, DE_NOVO, HDE, FIVE_TEN_K
```
zhx828, line 80: @calvinlu3 is it appropriate to name it FIVE_TEN_K ?

jhipster-jdl.jdl
```diff
@@ -76,34 +76,93 @@ entity CompanionDiagnosticDevice {
...
+}
+
+entity FdaSubmission {
```
zhx828, line 83: We still need to add the specimen type. But not sure whether it should be in FdaSubmission or CompanionDiagnosticDevice. Waiting for confirmation from Ritika.

jhipster-jdl.jdl
```diff
@@ -76,34 +76,93 @@ entity CompanionDiagnosticDevice {
...
+
+enum CopyNumberAlterationType {
+	AMPLIFICATION,
+    DELETION,
+    GAIN,
+    LOSS
+}
+
+enum StructuralVariantType {
+	FUSION
```
zhx828, line 113: Temporarily just add FUSION for now.

jhipster-jdl.jdl
calvinlu3: The FDA also uses `PMN` (Premarket Notification).
https://www.accessdata.fda.gov/scripts/cdrh/cfdocs/cfpmn/pmn.cfm?ID=K173492

zhx828: Yeah I saw that too. But the abbreviation seems only be used in the URL?

PR #57, opened 2022-04-06
Create Cdx and FdaSubmission parser

calvinlu3:
This is related to https://github.com/oncokb/oncokb/issues/3050

- Parsed the [CDx List](https://www.fda.gov/medical-devices/in-vitro-diagnostics/list-cleared-or-approved-companion-diagnostic-devices-in-vitro-and-imaging-tools) and stored the result as static data.


zhx828 (changes requested):

src/main/java/org/mskcc/oncokb/curation/util/CdxUtils.java
```diff
@@ -0,0 +1,227 @@
...
+
+    // Extract the relevant information from the PMA / 510(k) / HDE pages
+    public Set<FdaSubmission> getFDASubmissionFromHTML(Set<String> fdaSubmissionCodes) {
+        // Check if the fda submission codes are valid and purify input
+        Set<String> purifiedSubmissionCodes = new HashSet<>();
+        for (String code : fdaSubmissionCodes) {
+            code = code.toUpperCase();
+            if (code.matches("([A-Z]+[0-9]+)(\\/(S[0-9]+(-(S[0-9]+))?))?")) {
+                String[] splitCode = code.split("/");
+                String primaryPma = splitCode[0];
```
zhx828, line 92: You don't have to split if you use the match groups from line 90.

src/main/java/org/mskcc/oncokb/curation/util/CdxUtils.java
```diff
@@ -0,0 +1,227 @@
...
+                purifiedSubmissionCodes.add(primaryPma);
+                if (splitCode.length > 1) {
+                    String supplement = splitCode[1];
+                    String[] supplementRange = supplement.split("-");
```
zhx828, line 96: here as well, you can use match groups.

src/main/java/org/mskcc/oncokb/curation/util/CdxUtils.java
```diff
@@ -0,0 +1,227 @@
...
+                    fdaSubmission.setType(type.get());
+                }
+            }
+            fdaSubmissions.add(fdaSubmission);
+        }
+        return fdaSubmissions;
+    }
+
+    private Instant convertDateToInstant(String date) {
+        return LocalDate.parse(date, DateTimeFormatter.ofPattern("MM/dd/uuuu", Locale.US)).atStartOfDay().toInstant(ZoneOffset.UTC);
```
zhx828, line 201: This is interesting, I always use `yyyy`. Good to know! https://stackoverflow.com/questions/41177442/uuuu-versus-yyyy-in-datetimeformatter-formatting-pattern-codes-in-java

src/main/java/org/mskcc/oncokb/curation/util/CdxUtils.java
```diff
@@ -0,0 +1,227 @@
...
+
+            // Create CompanionDiagnosticDevice entity
+            CompanionDiagnosticDevice cdx = new CompanionDiagnosticDevice();
+            cdx.setName(tableCells.get(0));
+            Set<String> submissionCodes = Arrays
+                .asList(tableCells.get(1).split("\\s"))
+                .stream()
+                .map(code -> code.trim())
+                .filter(code -> code.length() > 0)
+                .map(code -> code.matches("DEN[0-9]{6}\\/K[0-9]{6}") ? code.split("/")[0] : code) // Special case for FerriScan where number is DEN130012/K124065
```
zhx828, line 75: I shall improve the matches. Maybe do a split and check whether the second element starts with `S`?

src/main/java/org/mskcc/oncokb/curation/util/CdxUtils.java
```diff
@@ -0,0 +1,227 @@
...
+                            purifiedSubmissionCodes.add(pmaString);
+                        });
+                }
+            }
+        }
+
+        // Fetch the PMA/510(k)/HDE information from webpage
+        Set<FdaSubmission> fdaSubmissions = new HashSet<>();
+        for (String code : purifiedSubmissionCodes) {
+            String prefix = code.split("[0-9]").length > 0 ? code.split("[0-9]")[0] : null;
```
zhx828, line 113: Other option is to use regex to get match group so it can be used directly without splitting twice.

src/main/java/org/mskcc/oncokb/curation/util/CdxUtils.java
```diff
@@ -0,0 +1,227 @@
...
+                        fdaSubmission.setNumber(content);
+                    } else if (header.equalsIgnoreCase("Generic Name")) {
+                        fdaSubmission.setGenericName(content);
+                    } else if (header.equalsIgnoreCase("Supplement Number")) {
+                        fdaSubmission.setSupplementNumber(content);
+                    } else if (header.equalsIgnoreCase("Date Received")) {
+                        fdaSubmission.setDateReceived(convertDateToInstant(content));
+                    } else if (header.equalsIgnoreCase("Decision Date")) {
+                        fdaSubmission.setDecisionDate(convertDateToInstant(content));
+                    }
```
zhx828, line 173: It would be better to use a switch to make the logic more clearer.

zhx828 (changes requested):

src/main/java/org/mskcc/oncokb/curation/util/CdxUtils.java
calvinlu3: I modified your regex slightly to `([A-Z]+[0-9]+)(\/((S[0-9]+)?(-(S[0-9]+))?))?` so that if we have a range (ie P050040/S001-S003), then groups 4 and 6 are S001 and S003, respectively.

src/main/java/org/mskcc/oncokb/curation/util/CdxUtils.java
calvinlu3: I believe that I can remove this line after switching to using match groups, as suggested. Match group 3 will be `null`, because regex is looking for S[0-9]+.

src/main/java/org/mskcc/oncokb/curation/util/CdxUtils.java
```diff
@@ -0,0 +1,230 @@
...
+            }
+            try {
+                rowSpanCount = Integer.parseInt(td.first().attributes().get("rowspan")) - 1;
+            } catch (NumberFormatException e) {}
+            ArrayList<String> tableCells = td.stream().limit(3).map(e -> e.text()).collect(Collectors.toCollection(ArrayList::new));
+
+            // Create CompanionDiagnosticDevice entity
+            CompanionDiagnosticDevice cdx = new CompanionDiagnosticDevice();
+            if (tableCells.size() < 3) {
+                continue; // Skip table cells with less than 3 elements
```
zhx828, line 72: Could you add a log.warn here specifying which line is skipped?

zhx828 (approved):

src/main/java/org/mskcc/oncokb/curation/util/CdxUtils.java
calvinlu3: Sure.

PR #59, opened 2022-04-13
Add endpoint to return parsed fda submissions

calvinlu3:
This is related to https://github.com/oncokb/oncokb/issues/3057

- For `FdaSubmission` entities, the `Number, Device Name, Companion Diagnostic Device, and Fda Submission Type ` are required
   - Also, cannot add new entity where `(number + supplement + cdx)` already exists
- Add endpoint at `/api/fda-submissions/lookup` to find an fda submission. If it doesn't already exist in db, then we use the parse the information from fda website
- Created `OncoKBTable.tsx`, which is a simple table using `react-table v7`
- Created our own [ValidatedForm](https://github.com/calvinlu3/oncokb-transcript/blob/fda-submission-endpoint/src/main/webapp/app/shared/form/ValidatedForm.tsx) and [ValidatedField](https://github.com/calvinlu3/oncokb-transcript/blob/fda-submission-endpoint/src/main/webapp/app/shared/form/ValidatedField.tsx), so form elements can be nested using `react-hook-forms`


src/main/webapp/app/entities/companion-diagnostic-device/companion-diagnostic-device.tsx
```diff
@@ -25,6 +27,35 @@ export const CompanionDiagnosticDevice = (props: ICompanionDiagnosticDeviceProps
 
   const { match } = props;
 
+  const columns: Column<ICompanionDiagnosticDevice>[] = [
+    { id: 'id', accessor: 'id', Header: 'ID' },
```
zhx828, line 31: Do we need to show the id if we don't allow clicking through the id to the device page like we used to do?

src/main/webapp/app/entities/fda-submission/fda-submission.store.ts
```diff
@@ -1,14 +1,33 @@
...
+    makeObservable(this, {
+      lookupFdaSubmission: action,
+    });
+  }
+
+  *lookupFdaSubmissionGen(submissionNumber: string) {
+    const numbers = submissionNumber.split('/');
+    const url = `${apiUrl}/lookup?number=${numbers[0]}${numbers[1] ? `&supplement_number=${numbers[1]}` : ''}`;
+    try {
+      const result: AxiosResponse<IFdaSubmission> = yield axios.get(url);
```
zhx828, line 24: Why do we use `yield` here?

src/main/webapp/app/entities/fda-submission/fda-submission.store.ts
```diff
@@ -1,14 +1,33 @@
...
+      this.reset();
+      this.entity = result.data ? result.data : {};
```
zhx828, line 26:
```suggestion
      this.entity = result.data || {};
```

src/main/webapp/app/entities/fda-submission/fda-submission.tsx
```diff
@@ -77,6 +79,63 @@ export const FdaSubmission = (props: IFdaSubmissionProps) => {
 
   const { match } = props;
 
+  const columns: Column<IFdaSubmission>[] = [
+    {
+      id: 'id',
+      accessor: 'id',
```
zhx828, line 85: I remember if the accessor is not specified, it uses id. Could you take a look? Though I do like list accessor always to make it clear.

src/main/webapp/app/shared/form/ValidatedForm.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { useEffect } from 'react';
+import { DefaultValues, FieldValues, FormProvider, SubmitHandler, useForm, useFormContext, ValidationMode } from 'react-hook-form';
+import { Form } from 'reactstrap';
+
+interface IValidatedFormProps {
+  children: React.ReactNode;
```
zhx828, line 6: You do not need to specify children.

zhx828 (changes requested):

src/main/webapp/app/entities/companion-diagnostic-device/companion-diagnostic-device.tsx
calvinlu3: We can omit that information. Also, just realized that for react-table columns, we don't need to `id` field if `accessor` is a string.

src/main/webapp/app/entities/fda-submission/fda-submission.tsx
calvinlu3: If the accessor is a string (not function), then `id` is not necessary. I will remove the `id` field.

src/main/webapp/app/entities/fda-submission/fda-submission.store.ts
calvinlu3: `generator-jhipster-react-mobx` uses [mobx flows](https://mobx.js.org/actions.html#asynchronous-actions), which takes in a generator function, for handling async methods.

src/main/webapp/app/config/constants.ts
```diff
@@ -32,3 +32,5 @@ export enum PAGE_ROUTE {
   OAUTH = '/oauth2/authorization/oidc',
   ADMIN_USER_MANAGEMENT = '/admin/user-management',
 }
+
+export const FDA_SUBMISSION_REGEX = new RegExp('^([A-Z]+[0-9]{6})(\\/((S[0-9]{3})(-(S[0-9]{3}))?))?');
```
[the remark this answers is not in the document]
calvinlu3: We will remove this restriction.

src/main/java/org/mskcc/oncokb/curation/service/FdaSubmissionService.java
```diff
@@ -99,6 +105,57 @@ public Optional<FdaSubmission> findOne(Long id) {
...
+     * @return Optional with the parsed fda submission or the one already existing in db, otherwise empty optional
+     */
+    public Optional<FdaSubmission> findOrFetchFdaSubmissionByNumber(String number, String supplementNumber) {
+        Optional<FdaSubmission> fdaSubmission = this.findByNumberAndSupplementNumber(number, supplementNumber);
+        if (fdaSubmission.isEmpty()) { // Fetch from FDA website if not present in database
+            if (StringUtils.isEmpty(number)) {
+                return Optional.empty();
+            }
+            String submissionNumber = number;
+            if (StringUtils.isEmpty(supplementNumber)) {
```
zhx828, line 131:
```suggestion
            if (!StringUtils.isEmpty(supplementNumber)) {
```

PR #70, opened 2022-05-17
Update DB models & Add protein change parser

zhx828:
- Update alteration entity
    - Add alteration_reference_genomes table for one-to-many mapping
    - remove alteration type
- Update consequence table
    - Add alteration type and name in the entity
- Rename the entity to Consequence from VariantConsequence
- Add alteration parser
- Update searchEngine option

@calvinlu3 don't worry about the commit histories. Ran into some issue when rebase, will squash before merging.


src/main/webapp/app/entities/alteration/alteration.store.ts
```diff
@@ -18,6 +21,15 @@ export class AlterationStore extends PaginationCrudStore<IAlteration> {
     this.totalItems = result.headers['x-total-count'];
     return this.entities;
   }
+  *annotateAlteration({ geneIds, alteration }) {
+    const alt = {
+      genes: geneIds.map(id => ({ id })),
+      alteration,
+    };
+    const result = yield new AlterationControllerApi(null, '', axiosInstance).annotateAlteration(alt as Alteration);
```
calvinlu3, line 29: I started to put some of the api clients in `shared/api/clients.ts`, so they can be reused. If that's a good idea, then we can move the `AlterationControllerAPI` client into that file as well.

calvinlu3 (approved):
LGTM

PR #71, opened 2022-05-25
Alteration parser updates

zhx828:
- Add reference allele when given position
- Parse fusion string properly
This fixes https://github.com/oncokb/oncokb/issues/2339

### DIFF
--- a/src/main/java/org/mskcc/oncokb/curation/util/AlterationUtils.java
+++ b/src/main/java/org/mskcc/oncokb/curation/util/AlterationUtils.java
@@ -14,6 +14,11 @@ import org.springframework.stereotype.Component;
 @Component
 public class AlterationUtils {
 
+    public static final String FUSION_SEPARATOR = "::";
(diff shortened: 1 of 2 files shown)


src/main/java/org/mskcc/oncokb/curation/util/AlterationUtils.java
```diff
@@ -296,8 +312,21 @@ public Alteration parseProteinChange(String proteinChange) {
...
+                genes.add(m.group(1));
+                genes.add(m.group(2));
+            } else {
+                p = Pattern.compile(FUSION_ALT_REGEX);
+                m = p.matcher(alteration);
+                if (m.matches()) {
+                    genes.add(m.group(1));
+                    genes.add(m.group(2));
+                }
+            }
```
calvinlu3, line 328: Group 2 could be an empty string if the input is something like `SYMBOL::`

calvinlu3 (approved):

PR #72, opened 2022-06-02
Updates to curation panel

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3136


src/main/webapp/app/shared/button/CurationModeButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+interface CurationModeButtonProps {
+  curateEntityId: string | number;
+}
+
+const CurationModeButton: React.FunctionComponent<
+  CurationModeButtonProps & StoreProps & ButtonProps & React.HTMLAttributes<HTMLButtonElement>
+> = props => {
+  const { toggleCurationPanel, curateEntityId, ...buttonProps } = props;
+  return (
+    <Button tag={Link} to={`/fda-submission/${curateEntityId}/curate`} color="secondary" type="button" {...buttonProps}>
```
zhx828, line 17: `/fda-submission/` should be passed through property. We could create entity enum and create route utils to return proper URL based on entity and action. Action should be enum as well.

zhx828 (changes requested):
- [ ] No need to uppercase `CURATION PANEL`
- [ ] Please check all places using SaveButton. Some of the properties are lost, especially the `disabled`

calvinlu3:
Ended up removing the `CurateButton` and merged the logic into `EntityActionButton` (which handles view, edit, curate, delete, and create).

zhx828 (approved):
Nicely done 💯

PR #76, opened 2022-06-13
Add elasticsearch reindex button

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3137

- Add `RouterPrompt.tsx` to display a modal when user tries to leave while ES is still indexing.
  - This component can be reused by any page that need to alert a user before they leave the page
- Add ElasticSearch management page to select the indices to reindex from a dropdown.

### DIFF
--- a/src/main/webapp/app/shared/prompt/RouterPrompt.tsx
+++ b/src/main/webapp/app/shared/prompt/RouterPrompt.tsx
@@ -1,0 +1,85 @@
+import { Location } from 'history';
+import React, { useEffect, useRef, useState } from 'react';
+import { Prompt, useHistory } from 'react-router-dom';
+import { Button, Modal, ModalBody, ModalFooter, ModalHeader } from 'reactstrap';
+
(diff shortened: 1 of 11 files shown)


src/main/webapp/app/shared/prompt/RouterPrompt.tsx
```diff
@@ -0,0 +1,85 @@
+import { Location } from 'history';
+import React, { useEffect, useRef, useState } from 'react';
+import { Prompt, useHistory } from 'react-router-dom';
+import { Button, Modal, ModalBody, ModalFooter, ModalHeader } from 'reactstrap';
+
+interface RouterPromptProps {
+  when: boolean | undefined;
```
zhx828, line 7: What does `when` mean here? It's a boolean field, I'm a confused.

zhx828 (changes requested):

zhx828 (approved):
👍🏼

PR #78, opened 2022-06-15
Update gene select dropdown to include aliases

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3161

 - Update curate button color
 
  image

- Add gene aliases to `GeneralSearch` gene options

image

- Add gene aliases to `GeneSelect` dropdown component

image

### DIFF
--- a/src/main/webapp/app/shared/select/GeneSelect.tsx
+++ b/src/main/webapp/app/shared/select/GeneSelect.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Props as SelectProps } from 'react-select';
+import { components, Props as SelectProps } from 'react-select';
 import { AsyncPaginate } from 'react-select-async-paginate';
 import { defaultAdditional } from 'app/components/curationPanel/FdaSubmissionPanel';
 import { DEFAULT_ENTITY_SORT_FIELD, DEFAULT_SORT_DIRECTION, ENTITY_TYPE, SearchOptionType } from 'app/config/constants';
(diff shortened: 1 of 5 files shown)


src/main/webapp/app/shared/select/GeneSelect.tsx
```diff
@@ -39,9 +40,29 @@ const GeneSelect = (props: IGeneSelectProps) => {
...
+      <>
+        <components.Option {...optionProps}>
+          <div>{optionProps.data.label}</div>
+          {hasAliases && (
+            <div style={{ color: 'grey', fontSize: '0.9em' }}>
+              <span>Also known as </span>
+              <span>{optionProps.data.geneAliases.map(alias => alias.name).join(', ')}</span>
+            </div>
+          )}
+        </components.Option>
```
zhx828, line 55: Can we move this to the shared folder so the gene and drug dropdowns or any dropdown option has subtitle? I believe the style will be the same and they should use the same component.

zhx828 (changes requested):

src/main/webapp/app/shared/select/SelectOption.tsx
```diff
@@ -0,0 +1,32 @@
...
+  label: string;
+  text: string;
+};
+
+interface IEntitySelectOption {
+  title: string;
+  subTitles?: SelectText[];
+}
+
+export const EntitySelectOption: React.FunctionComponent<IEntitySelectOption> = props => {
```
zhx828, line 13: Does it still support search keyword highlight now?

calvinlu3: No, the search keyword highlighting is available in the EntitySearchOption component: https://github.com/calvinlu3/oncokb-transcript/blob/add-alias-to-gene-select/src/main/webapp/app/shared/search/SearchOptions.tsx#L12-L44.

Should we also highlight keyword when we're searching in the gene or drug selects for example?

zhx828 (approved):

PR #80, opened 2022-07-13
Make wildcard query builder case insensitive

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/3169


calvinlu3 (approved):
Looks good, thanks for looking into this

PR #107, opened 2023-05-24
Support multiple spring security configs

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3458

### Changes
- There is now a security configuration that supports JWT token authentication that will be used by oncokb-core
- The other security  configuration is for OAuth2 that the frontend will be using
- The way the two configs are differentiated is by adding `requestMatcher(new RequestHeaderRequestMatcher("Authorization"))`
  - Only HTTP requests with Authorization header will use the JWT authentication method. Any other requests will use OAuth2

### DIFF
--- a/src/main/java/org/mskcc/oncokb/curation/config/security/SecurityConfigurationAPI.java
+++ b/src/main/java/org/mskcc/oncokb/curation/config/security/SecurityConfigurationAPI.java
@@ -1,0 +1,103 @@
+package org.mskcc.oncokb.curation.config.security;
+
+import org.mskcc.oncokb.curation.security.jwt.JWTConfigurer;
+import org.mskcc.oncokb.curation.security.jwt.TokenProvider;
+import org.springframework.context.annotation.Import;
+import org.springframework.core.annotation.Order;
+import org.springframework.http.HttpMethod;
+import org.springframework.security.config.annotation.method.configuration.EnableGlobalMethodSecurity;
+import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.config.annotation.web.builders.WebSecurity;
+import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
+import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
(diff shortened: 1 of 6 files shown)


src/main/java/org/mskcc/oncokb/curation/config/security/SecurityConfigurationAPI.java
```diff
@@ -0,0 +1,103 @@
...
+        .and()
+            .frameOptions()
+            .deny()
+        .and()
+            .sessionManagement()
+            .sessionCreationPolicy(SessionCreationPolicy.STATELESS)
+        .and()
+            .requestMatcher(new RequestHeaderRequestMatcher("Authorization"))
+            .authorizeRequests()
+            .antMatchers("/api/**").authenticated()
```
zhx828, line 92: We should have this clause at the very top of the configure. Most of other configurations probably not applicable.

src/main/java/org/mskcc/oncokb/curation/config/security/SecurityConfigurationOAuth.java
```diff
@@ -39,7 +39,7 @@ public class SecurityConfiguration extends WebSecurityConfigurerAdapter {
 
     private final SecurityProblemSupport problemSupport;
 
-    public SecurityConfiguration(
+    public SecurityConfigurationOAuth(
```
zhx828, line 42: We should add `@Order(2)` to clarify the logic.

zhx828 (changes requested):

src/main/java/org/mskcc/oncokb/curation/config/security/SecurityConfigurationAPI.java
calvinlu3: I moved the condition `.requestMatcher(new RequestHeaderRequestMatcher(JWTFilter.AUTHORIZATION_HEADER))` to the top, so its clear that this config is only for requests with Authorization header.
The `.antMatcher(/api/**) cannot be moved to top because both the JWT and OAuth security config should accept requests to api endpoints.

zhx828 (approved):

Issue #106, opened 2023-05-05
Look into using Firebase in oncokb-curation


calvinlu3:
We want to use Firebase in our new curation platform as well.

```[tasklist]
### Tasks
- [x] How to use Firebase in react
- [x] Support real-time editing
- [x] How to track history, so we can revert back to previous data
- [x] Firebase db rules and authentication
```

calvinlu3:
Here are two packages that look promising:

reactfire: https://github.com/FirebaseExtended/reactfire

react-firebase-hooks: https://github.com/CSFrequency/react-firebase-hooks

calvinlu3:
There is no package to help handle the history of the database, so we need to manually write some functions to keep track of changes.

PR #110, opened 2023-06-20
Configure Firebase in oncokb-curation

calvinlu3:
This fixes https://github.com/oncokb/oncokb-transcript/issues/106

Firebase Authentication and Authorization:
- Followed [guide](https://firebase.google.com/docs/auth/admin/create-custom-tokens#web-modular-api) on creating custom tokens for authentication. This option allows our server to perform the authentication. After successful authentication, then we can use `FirebaseAuth.getInstance().createCustomTokenAsync(email, additionalClaims)` to create a JWT token that the frontend will use to access firebase.
- These custom tokens are used purely for firebase auth, so frontend can access the protected database. 
- The `firebaseCustomToken` is added to the OAuth2AuthenticationToken object that is stored by Spring Security in sessions
- The frontend will call `api/account/firebase-token` to get the custom token to sign in when the app is initialized

Reading/Writing to Firebase from app:
- Installed `firebase-js-sdk` package to make changes to database
- Installed `react-firebase-hooks` package for consuming/reading from database


src/main/java/org/mskcc/oncokb/curation/security/CustomOAuthSuccessHandler.java
```diff
@@ -57,13 +61,33 @@ public void onAuthenticationSuccess(HttpServletRequest request, HttpServletRespo
...
-            redirectStrategy.sendRedirect(request, response, "/logout");
+
+            // Create custom token for frontend Firebase authentication
+            String email = authenticationWithAuthorities.getPrincipal().getAttribute("email");
+            String firebaseCustomToken = null;
+            try {
+                Map<String, Object> additionalClaims = new HashMap<>();
+                // This claim will be used in Firebase rules for authorization
+                additionalClaims.put(Constants.FIREBASE_AUTHORIZED_CLAIM, true);
+                firebaseCustomToken = FirebaseAuth.getInstance().createCustomTokenAsync(email, additionalClaims).get();
```
zhx828, line 72: This is a async method per the naming. Shouldn't wait for it?

src/main/java/org/mskcc/oncokb/curation/security/CustomOAuthSuccessHandler.java
```diff
@@ -57,13 +61,33 @@ public void onAuthenticationSuccess(HttpServletRequest request, HttpServletRespo
...
+            }
+
+            if (firebaseCustomToken != null) {
+                Map<String, Object> additionalDetails = new HashMap<>();
+                additionalDetails.put(Constants.FIREBASE_CUSTOM_TOKEN, firebaseCustomToken);
+                authenticationWithAuthorities.setDetails(additionalDetails);
+
+                SecurityContextHolder.getContext().setAuthentication(authenticationWithAuthorities);
+                super.onAuthenticationSuccess(request, response, authenticationWithAuthorities);
+                return;
```
zhx828, line 84: I assume the user will be logout if they don't have the access to the firebase? That's not the intended use case.

src/main/webapp/app/index.tsx
```diff
@@ -16,6 +19,10 @@ setupAxiosInterceptors(() => mobxStores.authStore.clearAuthentication('login.err
 
 loadIcons();
 
+// Initialize Firebase
+export const firebaseApp = initializeFirebase();
+export const firebaseDB = getDatabase(firebaseApp);
```
zhx828, line 24: If we do not check whether firebase configs are in place, we assume firebase is required. Then the oncokb-curation should not even be initialized(pod cannot be in ready state)

src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
```diff
@@ -75,6 +81,23 @@ public void importOncoKbSequence() throws ApiException, IOException {
...
+                        .getClassLoader()
+                        .getResource(applicationProperties.getFrontend().getFirebase().getServiceAccountCredentialsPath())
+                        .openStream()
+                )
+            )
+            .setDatabaseUrl(applicationProperties.getFrontend().getFirebase().getDatabaseUrl())
+            .build();
+
+        FirebaseApp.initializeApp(options);
+    }
```
zhx828, line 99: We use firebase in the backend, we should leave FirebaseProperties under applicationProperties and backfill the frontendProperties from applicationProperties

zhx828 (changes requested):

src/main/java/org/mskcc/oncokb/curation/security/CustomOAuthSuccessHandler.java
calvinlu3: The `.get()` waits for the token creation to complete and returns the result

src/main/java/org/mskcc/oncokb/curation/security/CustomOAuthSuccessHandler.java
calvinlu3: The user will be logged out if a firebase token cannot be created due to error. Should we still let the user login, but on the frontend we would need to make some components that rely on firebase unavailable?

zhx828: Yes, not all of users of curation platform will have the access to firebase.

zhx828: We could create a ROLE_FIREBASE to differentiate.

calvinlu3:
Main changes after code review:
- Added `FirebaseProperties` under `ApplicationProperties` as well. Added code to copy the properties added under `application.firebase` to `application.frontend.firebase`.
- Updated logic in `CustomOAuthSuccessHandler.java` to allow users without Firebase access to login by checking for `ROLE_FIREBASE` role. If they have that role, then we will add the firebase token as additional information in the Auth object.
- Move firebase related code to `firebase.store.ts`
- Converted `App.tsx` to functional component

zhx828 (approved):
LGTM, 👍🏼

PR #111, opened 2023-06-30
Fix some small issues with oncokb-curation

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3485

[Fix cancer types not showing up in dropdown](https://github.com/oncokb/oncokb-transcript/pull/111/commits/285d4e7cc387e2e6f148492ee67d07dc2731300d)
- We originally only searched by `mainType` so some cancer types did not show up.
- Support searching by `mainType, subtype, code`

[Make Firebase optional](https://github.com/oncokb/oncokb-transcript/pull/111/commits/2f8afee2e1a7ed6ec735f584cf316723e494ac9f)
- During development, we may not always be using Firebase but would still need to fetch the credentials. 

[Don't use Redis for Spring Sessions](https://github.com/oncokb/oncokb-transcript/pull/111/commits/4f724f325af9e2755273627f4350c6835285fd70)
- Set `spring.session.store-type` to `none`, so that Redis does not need to be started during dev

[Curation panel dropdown values not resetting](https://github.com/oncokb/oncokb-transcript/pull/111/commits/fbcb0c8fbca1a2605a288bed5123472d66e355b7)
- Re-render not triggered because we were using `display: none` to hide the panel, so state was not reset.

### DIFF
--- a/src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
+++ b/src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
@@ -92,19 +92,21 @@ public class OncokbCurationApp {
 
     @PostConstruct
     public void initFirebase() throws IOException {
-        FirebaseOptions options = new FirebaseOptions.Builder()
-            .setCredentials(
-                GoogleCredentials.fromStream(
-                    getClass()
-                        .getClassLoader()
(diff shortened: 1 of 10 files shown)


zhx828 (approved):
LGTM

PR #114, opened 2023-08-02
Create firebase crud store and add gene info section

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3499
This fixed https://github.com/oncokb/oncokb/issues/3498

### Changes
- Added classes based on existing Firebase models
- Created a Firebase CRUD store
- Added gene info section
image


src/main/webapp/app/shared/util/firebase-crud-store.ts
```diff
@@ -0,0 +1,85 @@
...
+    set(ref(this.db, path), value);
+  }
+
+  push(path: string, value: any) {
+    const listRef = ref(this.db, path);
+    const newItemRef = push(listRef);
+    set(newItemRef, value);
+  }
+
+  update(path: string, value: RecursivePartial<T>) {
```
calvinlu3, line 75: I added `value: RecursivePartial<T>`, so we can have some type hints when updating.

calvinlu3:
## Changes Requested
1. The input fields should do more than changing the content. (ie. Changing the background summary should also trigger a change in lastReview, lastUpdate author, meta collection edit, etc). These should be done through the store or through firebase directly. 
2. See `realtimeString.js` in old curation platform. We can create a functional/class component for all the different types of inputs.
3. We should think about how we want to embed the review view
4. Fix async useEffect in `gene-curate.tsx`
5. In `gene-curate`, consolidate the `on[field]Change` handlers. The curate page will have a lot of these handlers, so we need to have a generic way of updating.


## Realtime Editing Gene Side Effects
Currently, we are using `DefaultField`, which is a simple input that updates to Firebase. However, there are actually additional changes to other collections that we may need to update when updating a field.

1. After updating a field to Firebase, we also need to:
      - `Meta.gene.lastModifiedBy` and `Meta.gene.lastModifiedAt` should be updated. This information is shown at the top of the gene page.
      - `Meta.gene.[field]_review` object should be updated. For instance, if `background` is updated, then `background_review` should also be updated. This object includes `updateTime`, `updatedBy`, `lastReviewed`.
      - `Meta.gene.[field]_editing` should be updated. This is used to stop multiple users from editing same input.
      - `Meta.gene.review.uuid` should be set to true. This is the uuid of the fields that need to be reviewed for this gene.
2. When user enters the gene page, the gene name should be added to array `Meta.collaborators.name[n]`. This is used to store the users currently working on the gene.

3. When user enters review mode:
     - `Meta.gene.review.currentReviewer` should be set to current user's name

src/main/webapp/app/shared/util/firebase-crud-store.ts
```diff
@@ -1,7 +1,8 @@
 import { IRootStore } from 'app/stores';
```
zhx828, line 1: This file should probably be located under the firebase folder?

src/main/webapp/app/shared/firebase/FirebaseRealtimeInput.tsx
```diff
@@ -0,0 +1,48 @@
...
+  label?: string;
+  inputClass?: string;
+}
+
+export const RealtimeBasicInput: React.FunctionComponent<IRealtimeBasicInput> = props => {
+  const { name, id = name, type, onChange, className, labelClass, label, inputClass, children, ...otherProps } = props;
+  return (
+    <div className={`mb-2 ${className}`}>
+      {label && (
+        <Label id={name} for={name} className={labelClass}>
```
zhx828, line 28: `<Label id={id} for={id} className={labelClass}>`

src/main/webapp/app/shared/firebase/FirebaseRealtimeInput.tsx
```diff
@@ -0,0 +1,48 @@
...
+      )}
+      <Input
+        className={inputClass}
+        id={id}
+        name={name}
+        autoComplete="off"
+        onChange={e => {
+          onChange && onChange(e);
+        }}
+        type={props.type as any}
```
zhx828, line 40: Can we do ` type={props.type as string}`?

src/main/webapp/app/shared/firebase/FirebaseRealtimeInput.tsx
```diff
@@ -0,0 +1,48 @@
...
+        style={type === 'checkbox' ? { marginLeft: '0.25rem' } : null}
```
zhx828, line 41: Should the label on the right when it's checkbox/radio?

src/main/webapp/app/entities/gene/gene-curate.tsx
```diff
@@ -108,76 +95,74 @@ export const GeneCurate = (props: IGeneCurateProps) => {
...
                 labelClass="font-weight-bold"
                 name="geneBackground"
-                type="textarea"
+                type={RealtimeInputType.TEXTAREA}
                 value={props.data?.background || ''}
-                rows={Math.ceil(props.data?.background.length / 250)}
-                onChange={event => {
-                  props.update(firebaseParentPath, { background: event.target.value });
+                onChange={e => {
+                  props.updateReviewableContent(firebaseParentPath, 'background', e.target.value);
```
zhx828, line 165: We should have the path written in constants too. Even better if we could have an util to generate the path.

zhx828 (commented):
Looks good!

src/main/webapp/app/config/constants.ts
```diff
@@ -115,4 +122,30 @@ export const DEFAULT_ENTITY_SORT_FIELD: { [key in ENTITY_TYPE]?: string } = {
...
+  HISTORY = 'History',
+  MAP = 'Map',
+  META = 'Meta',
+  QUEUES = 'Queues',
+  SETTING = 'Setting',
+  USERS = 'Users',
+  VUS = 'VUS',
+}
+
+export const FB_COLLECTION_PATH = {
```
calvinlu3, line 142: Added constants for firebase paths

src/main/webapp/app/pages/curation/GeneListPage.tsx
```diff
@@ -0,0 +1,174 @@
+import React, { useEffect, useMemo, useState } from 'react';
```
calvinlu3, line 1: This is new and needs review.

Also note that we will moving the pagination/searching functionality of the table to the table component in a future PR (track by ticket: https://github.com/oncokb/oncokb/issues/3564)

src/main/webapp/app/routes.tsx
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
```
calvinlu3, line 1: Moved the curation page out of the gene entity page

src/main/webapp/app/shared/firebase/FirebaseRealtimeInput.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const inputComponent = (
+    <Input
+      className={classNames(inputClass, isCheckType && 'ml-1 position-relative')}
+      id={id}
+      name={name}
+      autoComplete="off"
+      onChange={e => {
+        onChange && onChange(e);
+      }}
+      type={props.type as InputType}
```
calvinlu3, line 45: Change from `as any` to `as InputType`

src/main/webapp/app/shared/firebase/FirebaseRealtimeInput.tsx
```diff
@@ -0,0 +1,68 @@
...
+          {inputComponent}
+          {labelComponent}
+        </>
+      ) : (
+        <>
+          {labelComponent}
+          {inputComponent}
+        </>
+      )}
+    </div>
```
calvinlu3, line 66: Change the label position depending on if the input is a checbox/radio

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -0,0 +1,182 @@
...
+                }}
+              />
+              <div className="mb-2">
+                <AutoParseRefField summary={props.data?.background} />
+              </div>
+              <div>
+                <span className="font-weight-bold mr-2">External Links:</span>
+                <WithSeparator separator={InlineDivider}>
+                  <a href={`https://cbioportal.mskcc.org/ln?q=${props.data?.name}`} target="_blank" rel="noopener noreferrer">
+                    CBioPortal <ExternalLinkIcon />
```
zhx828, line 144:
```suggestion
                    cBioPortal <ExternalLinkIcon />
```

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -0,0 +1,182 @@
...
+                  </a>
+                  <a
+                    href={`http://cancer.sanger.ac.uk/cosmic/gene/overview?ln=${props.data?.name}`}
+                    target="_blank"
+                    rel="noopener noreferrer"
+                  >
+                    Cosmic <ExternalLinkIcon />
```
zhx828, line 151:
```suggestion
                    COSMIC <ExternalLinkIcon />
```

zhx828 (changes requested):

src/main/webapp/app/pages/curation/GeneListPage.tsx
calvinlu3: `OncokbTable` from the public repo does have sorting/pagination, but the table component in this repo does not. We are using react-table-v7 and public is using react-table-v6, so we can't completely copy over the table.

@zhx828 Do you think we should downgrade to v6 and just reuse the table from public repo?

I agree that we should have two table components, one for server-side and one for client side pagination/search. I will update the ticket

src/main/webapp/app/oncokb-commons/components/RefComponent.tsx
```diff
@@ -0,0 +1,62 @@
+import { getNCBIlink } from './lib/urls';
+
+export type ParsedRef = {
+  prefix?: string;
+  content: string;
+  link?: string;
+  isAbstract?: boolean;
+};
+
+export const parseReferences = (content: string, seperatePMIDs = false): ParsedRef[] => {
```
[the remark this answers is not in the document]
calvinlu3: I am trying to extract the logic from [here](https://github.com/oncokb/oncokb-public/blob/master/src/main/webapp/app/oncokb-frontend-commons/src/components/RefComponent.tsx) to a function. The code from oncokb-public frontend commons returns a component, but I just need to parsing logic for my use case.

I think I should probably move this to a util file instead of under `oncokb-commons` folder.

src/main/webapp/app/shared/util/url-utils.ts
```diff
@@ -0,0 +1,14 @@
+export const replaceUrlParams = (url: string, ...params: string[]) => {
```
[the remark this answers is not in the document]
calvinlu3: Added some tests in this [commit](https://github.com/oncokb/oncokb-transcript/pull/114/commits/c7a288ec3c60b7baf420fb6fa75b5fec20e099ea).

The function will try to replace as many params in the url.
  - If we pass in more params than the url has, then it will ignore the extras
  - If we pass in fewer params than the url, then the missing params won't be replaced in url

src/main/webapp/app/shared/util/firebase/firebase-crud-store.ts
```diff
@@ -0,0 +1,120 @@
...
+/* eslint-disable @typescript-eslint/ban-types */
+export type RecursiveKeyOf<TObj extends object> = {
+  [TKey in keyof TObj & (string | number)]: TObj[TKey] extends any[]
+    ? `${TKey}`
+    : TObj[TKey] extends object
+    ? `${TKey}` | `${TKey}/${RecursiveKeyOf<TObj[TKey]>}`
+    : `${TKey}`;
+}[keyof TObj & (string | number)];
+
+export class FirebaseCrudStore<T> {
```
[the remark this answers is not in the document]
calvinlu3: I'm trying to follow based on how JHipster generated `base-crud-store.ts`, `base-store.ts`, etc, which are all located under the utils folder. 

I think it would make sense if we move all `stores` to `webapp/stores` then?

src/main/webapp/app/pages/curation/CurationPage.tsx
calvinlu3: Added `cBioPortal` and `COSMIC` strings to the `constants.ts` file.

src/main/webapp/app/pages/curation/GeneListPage.tsx
zhx828: Not familiar with v7. Any reason we upgraded?

src/main/webapp/app/shared/util/url-utils.ts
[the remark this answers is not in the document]
zhx828: I think the last scenario should resolve in exception. Practical way is to return an empty url and handle at the place where this method is called. Or create a component to redirect to main page when rerun is an empty URL's. A url without all parameters will be an invalid url anyway, it should be redirected to the main page with a warning banner?

src/main/webapp/app/shared/util/firebase/firebase-utils.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const getValueByNestedKey = (obj: any, nestedKey = '') => {
+  return nestedKey.split('/').reduce((currObj, currKey) => {
+    if (currObj) return currObj[currKey];
+  }, obj);
+};
+
+export const geneNeedsReview = (meta: Meta) => {
+  let needsReview = false;
+  const metaReview = meta?.review;
```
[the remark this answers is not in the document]
zhx828: Then let's change the param type by including the undefined

src/main/webapp/app/shared/util/firebase/firebase-crud-store.ts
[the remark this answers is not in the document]
zhx828: I think they should all be included under the firebase folder. Let's discuss during the standup

PR #119, opened 2023-08-18
Merge oncokb-curation feature branch into master

calvinlu3:


calvinlu3:
Running into an issue where `geneAliases` field in `Gene` schema is not present in the api-docs

PR #124, opened 2023-09-26
Fix firebase auth error

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3561

### Cause of issue
When the firebase custom token is minted in the backend, it has an expiration time of 1 hour. This means that our frontend application has 1 hour to use that token in the `signInWithCustomToken` method from `firebase/auth` library. 

1. When we refresh the page, we try to `initializeFirebase()` which ultimately calls the  `signInWithCustomToken()`. So, every time the page is refreshed, we try to use the custom token to sign in again. Since the custom token has an expiration of 1 hour, it cannot be used to sign in after 1 hour. 

2. The second way we can get this error is if the user has been idle for >30mins. By default, spring session has a max idle timeout of 30 minutes, so after that, it would try to get the user to relog. When this happens, the user is redirected to keycloak page and upon successful login, they are redirected back to the application. This process triggers a refresh.

### Solution
After we call `signInWithCustomToken`, Firebase SDK will handle refreshing the token for us. By default, firebase stores our user credentials inside `IndexDB` and use `LOCAL` persistence. ([Ref](https://firebase.google.com/docs/auth/web/auth-state-persistence#:~:text=firebase.auth.Auth.Persistence.LOCAL)) This means that the user should remain signed in even after page is refreshed, browser is closed, opens new tab until they are explicitly signed out.

image

All we have to do is create the listener `onAuthStateChanged` to only sign in with custom token when firebase cannot find the user credentials in `IndexDB`.

### DIFF
--- a/src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
+++ b/src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
@@ -5,6 +5,7 @@ import com.google.firebase.FirebaseApp;
 import com.google.firebase.FirebaseOptions;
 import java.io.IOException;
 import java.net.InetAddress;
+import java.net.URL;
 import java.net.UnknownHostException;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


zhx828 (approved):
LGTM, thanks!

PR #128, opened 2023-10-17
Replace search endpoints

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3580


zhx828 (approved):
Looks good. Should we run the frontend script to update the API clients?

PR #129, opened 2023-10-31
Fix curation page stuck in loading for certain genes

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3594

- Update search url to fix infinite loading for certain genes
- Check if mutations exists before rendering
- Add actions column to user table


zhx828 (approved):
LGTM

PR #133, opened 2023-11-08
Add ROLE_CURATOR and update navigation panel according to user role

calvinlu3:
This is related to https://github.com/oncokb/oncokb/issues/3596


zhx828 (approved):
LGTM, thanks.

PR #136, opened 2023-11-14
Incorporate JaVers as audit framework & Remove History entity

zhx828:
This fixes https://github.com/oncokb/oncokb/issues/3604
This is related to https://github.com/oncokb/oncokb/issues/3597


src/main/java/org/mskcc/oncokb/curation/web/rest/AuditResource.java
```diff
@@ -0,0 +1,131 @@
...
+        @RequestParam(value = "entityId") Integer entityId,
+        @RequestParam(value = "commitId") Long commitId
+    ) throws ClassNotFoundException {
+        Class entityTypeToFetch = Class.forName(ENTITY_PACKAGE_PATH + entityType);
+
+        // Find the commit
+        QueryBuilder jqlQuery = QueryBuilder.byInstanceId(entityId, entityTypeToFetch).withCommitId(BigDecimal.valueOf(commitId)).limit(5);
+
+        // Since we use random commit id, we need to find the list of commits that prior to the commitId
+        List<CdoSnapshot> snapshots = javers.findSnapshots(jqlQuery.build());
```
calvinlu3, line 116: Are commit ids unique? I'm wondering why we need the limit 5. What is the difference if I set to limit 1?

src/main/java/org/mskcc/oncokb/curation/repository/AlterationRepository.java
```diff
@@ -12,6 +13,7 @@
 /**
  * Spring Data SQL repository for the Alteration entity.
  */
+@JaversSpringDataAuditable
```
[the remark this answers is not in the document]
zhx828: @calvinlu3 I can't use blueprint unfortunately. I had to add manually, but it was straightforward. It has compatibility issue with our jhipster version.

`@JaversSpringDataAuditable` is specifically designed for Repository to attach to CRUD.

src/main/java/org/mskcc/oncokb/curation/web/rest/AuditResource.java
zhx828: Yeah, makes sense. Must be a legacy code I forgot to update.

PR #139, opened 2023-11-20
Add vus table

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3607
resolves https://github.com/oncokb/oncokb/issues/3611
resolves https://github.com/oncokb/oncokb-transcript/issues/141


src/main/webapp/app/components/commentIcon/CommentIcon.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+function CommentIcon(props: ICommentIconProps) {
+  let color: string;
+  if (props.comments.length === 0) {
+    color = 'black';
+  } else if (props.comments.some(comment => comment.resolved === 'false')) {
+    color = 'gold';
+  } else {
+    color = 'green';
+  }
```
zhx828, line 27: The color should be a computed value so when comments change, the color will be updated.

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
```diff
@@ -0,0 +1,160 @@
...
+      setShowComments(false);
+    }, CLOSE_POPOVER_DURATION_MILLISECONDS);
+
+    setTimeoutId(id);
+  }
+
+  return (
+    <div style={{ display: 'inline-block' }} onMouseEnter={handleMouseEnter} onMouseLeave={handleMouseLeave}>
+      <FontAwesomeIcon id="commentIcon" icon={faComment} color={color} />
+      <Popover target="commentIcon" isOpen={showComments}>
```
zhx828, line 49: Did you check out the rc-tooltip?

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
```diff
@@ -0,0 +1,160 @@
...
+    };
+  }, [isCounting, timeRemainingUntilDelete]);
+
+  useEffect(() => {
+    return () => {
+      if (syncedIsCounting.current) {
+        props.onDeleteComment();
+      }
+    };
+  }, []);
```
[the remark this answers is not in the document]
bprize15: Yeah this is the hacky part. I need to delete all comments currently deleting when the component unmount, so that's why this useEffect exists. I need to use to maintain a version of the isCounting state that is up to date with the current render cycle. Since the useEffect cleanup function is meant to clean up effects from the previous render, the state will be stale if I do not maintain this ref

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
bprize15: Have not yet, mostly just wanted to get the functionality working, but should be an easy replace if we decide to go with tooltip

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
```diff
@@ -0,0 +1,160 @@
...
+        setTimeRemainingUntilDelete(time => time - 1);
+      }, 1000);
+    } else if (isCounting && timeRemainingUntilDelete === 0) {
+      props.onDeleteComment();
+      setIsCounting(false);
+      setTimeRemainingUntilDelete(10);
+    }
+
+    return () => {
+      clearTimeout(timer);
```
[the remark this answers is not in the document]
bprize15: By delete do you mean unmount? If, so the second useEffect handles when the component unmount. I just need this useEffect to make the timer count down from 10, and delete the comment when it reaches 0. Since we will be listening to firebase to changes from comments, when the list of comments changes in firebase, that should trigger a rerender, and the deleted comment should disappear

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
bprize15: I believe this will automatically update the color? Since the VUS listener is a state, when the comments change, that should trigger a rerender which should rerun this logic.

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
```diff
@@ -0,0 +1,245 @@
...
+  onCreateComment: (content: string) => void;
+  onResolveComment: (index: number) => void;
+  onUnresolveComment: (index: number) => void;
+  onDeleteComments: (indices: number[]) => void;
+}
+
+const CommentBox = observer((props: ICommentBoxProps) => {
+  const scrollContainerRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
```
calvinlu3, line 92: How come these useEffects have no dependency arrays?

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
```diff
@@ -0,0 +1,245 @@
...
+      }
+    };
+  });
+
+  const sortedCommentItems = props.comments
+    ?.map((comment, index) => {
+      return [
+        comment.date,
+        <CommentItem
+          key={`${comment.date}-${comment.email}-${comment.content}`}
```
calvinlu3, line 113: I think date and email is sufficient as a unique id?

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
```diff
@@ -0,0 +1,245 @@
...
+    .sort((comment1, comment2) => {
+      return parseInt(comment2[0] as string, 10) - parseInt(comment1[0] as string, 10);
+    })
+    .map(comment => comment[1]);
+
+  return (
+    <>
+      {sortedCommentItems && sortedCommentItems.length > 0 && (
+        <div ref={scrollContainerRef} style={{ maxHeight: '320px', overflow: 'auto' }}>
+          <ListGroup className="ml-3 mr-3 mb-3 mt-3" flush>
```
calvinlu3, line 136: I think you can do this with bootstrap to specific all sides? Not 100% certain
```suggestion
          <ListGroup className="m-3" flush>
```

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -302,15 +305,32 @@ const CurationPage = (props: ICurationPageProps) => {
...
+        </div>
+      )}
+      {props.vusData && (
+        <VusTable
+          vusObjList={props.vusData}
+          hugoSymbol={hugoSymbol}
+          account={props.account}
+          handleFirebaseUpdate={props.updateVus}
+          handleFirebaseDelete={props.deleteVus}
+          handleFirebaseCreateAnyValue={props.createAnyValueVus}
```
calvinlu3, line 317: I think you can move the store props down from CurationPage to the VusTable component. You would need to use `componentInject` on the VusTable component. Something like below:

```
const VusTable = (props) => {...}

const mapStoreToProps = ({ firebaseVusStore }: IRootStore) => ({
  vusData: firebaseVusStore.data,
  addVusListener: firebaseVusStore.addListener,
  ...
});

type StoreProps = Partial<ReturnType<typeof mapStoreToProps>>;

export default componentInject(mapStoreToProps)(observer(VusTable));
```

calvinlu3: We use `Partial` for the type because we want to make all the store props as optional because mobx will inject it for us. Without the optional, whenever you use `VusTable` it will complain that you need to pass in the props.

src/main/webapp/app/pages/curation/VusTable.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+  return (
+    <div className={'justify-content-between align-items-center mt-4'}>
+      <Col>
+        <Row className="justify-content-between mb-3">
+          <h5>Variants of Unknown Signficance (Investigated and data not found):</h5>
+          <div>
+            <Button onClick={handleUpdate} className="mr-1" color="primary" size="md">
+              Update to database
+            </Button>
```
calvinlu3, line 234: Let's remove this button and the handler for now

src/main/webapp/app/shared/modal/SimpleConfirmModal.tsx
```diff
@@ -17,19 +25,38 @@ export const SimpleConfirmModal: React.FunctionComponent<{
...
+            </>
+          )}{' '}
+          {props.confirmText || 'Confirm'}
         </Button>
       </ModalFooter>
     </Modal>
   );
 };
+
+// <Button color="secondary" onClick={handleClose}>
```
calvinlu3, line 55: This can be removed?

src/main/webapp/app/stores/open-comment.store.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  clear() {
+    this.commentIndiciesToDelete = [];
+  }
+}
+
+export const openCommentScrollPosition = observable.box<number>(null);
+export const openCommentsIdStore = observable.box<string>(null);
+export const commentsToDeleteStore = new CommentsToDeleteStore();
```
calvinlu3, line 30: I think you can initiate the store in the `createStore.ts` folder where all our stores are defined

src/main/webapp/app/stores/open-comment.store.ts
```diff
@@ -0,0 +1,30 @@
...
+    _.remove(this.commentIndiciesToDelete, commentIndex => commentIndex === index);
+  }
+
+  clear() {
+    this.commentIndiciesToDelete = [];
+  }
+}
+
+export const openCommentScrollPosition = observable.box<number>(null);
+export const openCommentsIdStore = observable.box<string>(null);
```
calvinlu3, line 29: Wondering why these are not in the comment store and as constants?

src/main/webapp/app/stores/open-comment.store.ts
calvinlu3: If you instantiate this store in the createStore, then you can directly inject this store into the component that you need to use the data in.

You can use the same approach as described in my comment about VusTable

calvinlu3 (changes requested):

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
bprize15: I want the effect to run on every rerender, not just when the component mounts. As for the second one, I only want it to run on unmount, so I think I can just add an empty dependency array.

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
bprize15: You're right. Thanks for catching this

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
```diff
@@ -0,0 +1,245 @@
...
+      }
+    }, CLOSE_TOOLTIP_DURATION_MILLISECONDS);
+
+    timeoutId.current = id;
+  }
+
+  return (
+    <div onMouseEnter={handleMouseEnter} onMouseLeave={handleMouseLeave}>
+      <DefaultTooltip
+        overlayStyle={{ opacity: 1 }}
```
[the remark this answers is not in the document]
bprize15: Done

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
```diff
@@ -0,0 +1,245 @@
...
+            comments={props.comments}
+            onCreateComment={props.onCreateComment}
+            onResolveComment={props.onResolveComment}
+            onUnresolveComment={props.onUnresolveComment}
+            onDeleteComments={props.onDeleteComments}
+          />
+        }
+      >
+        <div>
+          <FontAwesomeIcon size={props.size || 'lg'} id={props.id} icon={faComment} color={color} />
```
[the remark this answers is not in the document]
bprize15: They seem to look the exact same. Will leave as is unless you think it's more consistent to use the react icons

src/main/webapp/app/stores/open-comment.store.ts
bprize15: FIxed

zhx828 (changes requested):
- We either should move the comment icon into the Action column, or come up with a different visualization. Aligning on the right of the Latest Comment doesn't look very appealing, especially when there is no comment for all the alterations within the page.
- There is a lot of space under the VUS header. Can we try to move the Download button next to the header and move the search to the same line of the title?
- When last comment is too long, we should use ellipsis text 
Screenshot 2023-12-04 at 10 24 54 PM

- The table should be sorted by last edit ascending order 
- The sorting icon is off in my local version, do you see the same thing?  Screenshot 2023-12-04 at 10 19 43 PM

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
```diff
@@ -0,0 +1,260 @@
...
+}
+
+function CommentInput(props: ICommentInputProps) {
+  const [commentInputValue, setCommentInputValue] = useState('');
+
+  function handleKeyPress(event: React.KeyboardEvent<HTMLInputElement>) {
+    if (commentInputValue !== '' && event.key === 'Enter') {
+      props.onCreateComment(commentInputValue);
+    }
+  }
```
zhx828, line 170: 👍🏼  thanks for adding. I do expect this to be a default behavior for input group with a primary action button components. cc @calvinlu3 
Not need to fix in this PR, we will need to do it in the next feature where the input group needed.

src/main/webapp/app/shared/table/VusTable.tsx
```diff
@@ -0,0 +1,264 @@
...
+                <FontAwesomeIcon icon={'trash'} />
+              </Button>
+            </Row>
+          </Container>
+        );
+      },
+    },
+  ];
+
+  return vusData ? (
```
zhx828, line 218: Do we have a ticket on allowing user to add VUS? If not, we need to add that in another PR.
At least the header of the table should be available to all genes no matter vus available.

src/main/webapp/app/shared/table/VusTable.tsx
```diff
@@ -0,0 +1,264 @@
...
+        <h5 className="mb-3">Variants of Unknown Signficance (Investigated and data not found):</h5>
+        <div>
+          <Button onClick={handleDownload} color="primary" size="md">
+            Download
+          </Button>
+        </div>
+      </div>
+      <OncoKBTable defaultSorted={[{ id: LAST_EDITED_AT, desc: false }]} data={vusList} columns={columns} showPagination />
+      <SimpleConfirmModal
+        title="Confirm delete operation"
```
zhx828, line 230:
```suggestion
        title="Confirm deletion"
```

src/main/webapp/app/shared/table/VusTable.tsx
```diff
@@ -0,0 +1,264 @@
...
+        body={'Are you sure you want to delete this VUS?'}
```
zhx828, line 231:
```suggestion
        body={'Are you sure you want to delete this variant?'}
```

src/main/webapp/app/shared/util/utils.tsx
```diff
@@ -116,3 +118,36 @@ export const getEntityTableActionsColumn = (entityType: ENTITY_TYPE) => {
...
+
+export function getMostRecentComment(comments: Comment[]) {
+  let latestComment = comments[0];
+  for (const comment of comments) {
+    if (parseInt(comment.date, 10) > parseInt(latestComment.date, 10)) {
+      latestComment = comment;
+    }
+  }
+  return latestComment;
+}
```
zhx828, line 142: This should be moved to firebase-utils.ts, let's create it if not available. cc @calvinlu3

zhx828 (changes requested):
- One thing I noticed is that over the changes, you use the useRef over useState. Any thoughts on making that decision?
- How do we handle data update exception? For instance, for all handle events, we assume data will be updated successfully. We should at least show error notification when that fails. cc @calvinlu3

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
```diff
@@ -0,0 +1,260 @@
...
+        </div>
+      </DefaultTooltip>
+    </div>
+  );
+});
+
+export interface ICommentBoxProps {
+  commentStore: CommentStore;
+  comments: Comment[];
+  openCommentId: string;
```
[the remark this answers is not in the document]
bprize15: It handles setting the correct comment box to be open when the page rerenders

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
```diff
@@ -0,0 +1,260 @@
...
+  function handleKeyPress(event: React.KeyboardEvent<HTMLInputElement>) {
+    if (commentInputValue !== '' && event.key === 'Enter') {
+      props.onCreateComment(commentInputValue);
+    }
+  }
+
+  return (
+    <Container>
+      <Row className="border-top rounded-lg bg-light">
+        <InputGroup className="m-3">
```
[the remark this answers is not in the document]
bprize15: I think the border separating the comments looks a little weird if it's the same color as the comment box border. I think it looks much better lighter

src/main/webapp/app/components/commentIcon/CommentIcon.tsx
```diff
@@ -0,0 +1,260 @@
...
+  useEffect(() => {
+    const scrollPosition = props.commentStore.openCommentsScrollPosition;
+    if (scrollPosition && scrollContainerRef.current) {
+      scrollContainerRef.current.scrollTop = scrollPosition;
+    }
+  });
+
+  useEffect(() => {
+    return () => {
+      if (props.comments && props.commentStore.openCommentsId !== props.openCommentId) {
```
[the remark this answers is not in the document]
bprize15: I think it makes more sense to make them both `openCommentsId` since it refers to the id that contains all open comments

src/main/webapp/app/stores/firebase/firebase.comment.store.ts
```diff
@@ -0,0 +1,39 @@
...
+      removeCommentToDeleteIndex: action.bound,
+      clearCommentsToDelete: action.bound,
+    });
+  }
+
+  addCommentToDeleteIndex(index: number) {
+    this.commentIndiciesToDelete.push(index);
+  }
+
+  removeCommentToDeleteIndex(index: number) {
```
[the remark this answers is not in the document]
bprize15: It's supposed to be saying "I want to remove a comment's index from the list of indexes that should be deleted." Maybe just `removeIndexToDelete`?

src/main/webapp/app/shared/table/VusTable.tsx
bprize15: We do not yet. Maybe we can discuss in standup and make that edit when we add the 'Add VUS` button?

zhx828 (approved):
LGTM

PR #140, opened 2023-11-28
Add mutation and cancer type level summaries to Collapsible header

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3606

Here are some examples of how it looks like:
image

image

### DIFF
--- a/src/main/webapp/app/shared/badge/CountBadge.tsx
+++ b/src/main/webapp/app/shared/badge/CountBadge.tsx
@@ -15,7 +15,7 @@ const CountBadge = (props: CountBadgeProps) => {
 
   return (
     <div className="mr-1">
-      <div className={classNames('pt-2', hideWhenOne ? 'pr-0' : 'pr-2', 'count-badge-wrapper')}>
+      <div className={classNames('pt-2', 'pr-2', 'count-badge-wrapper')}>
         <span className="font-weight-bold">{props.base}</span>
         {hideWhenOne ? undefined : (
           <span className="count-badge">
--- a/src/main/webapp/app/shared/badge/CountBadge.tsx
+++ b/src/main/webapp/app/shared/badge/CountBadge.tsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import './count-badge.scss';
+import classNames from 'classnames';
+
+export interface CountBadgeProps {
+  base: string | JSX.Element;
+  count: number;
+  hideWhenOne?: boolean;
+  max?: number;
+}
+
+const CountBadge = (props: CountBadgeProps) => {
+  const count = props.count > (props.max || Number.MAX_SAFE_INTEGER) ? `${props.max}+` : props.count;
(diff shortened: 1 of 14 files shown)


src/main/webapp/app/shared/badge/count-badge.scss
```diff
@@ -0,0 +1,25 @@
+.oncokb.icon {
+  margin: 0;
+}
+
+.count-badge-wrapper {
+  font-size: 90%;
```
calvinlu3, line 6: Is there a way to shrink the size of the oncokb icons from oncokb-styles? I can make the text's font size smaller, if I can adjust the icon's size, but I don't see a way to do that.

Another option is that I can make the mutation name and all other bolded text `h4`, but keep the default font-size for all other text.

webpack/webpack.common.js
```diff
@@ -68,12 +68,7 @@ module.exports = async options => {
           },
           {
             test: /\.(jpe?g|png|gif|svg|woff2?|otf|ttf|eot|ppt|pdf|zip)$/i,
-            loader: 'file-loader',
-            options: {
-              digest: 'hex',
-              hash: 'sha512',
-              name: 'content/[name].[ext]',
-            },
+            type: 'asset/resource',
```
calvinlu3, line 71: This change resolved the oncokb-style icons not loading. I originally copied over the webpack config from oncokb-public, but that project uses Webpack v4, while we use Webpack v5.
Reference: https://webpack.js.org/guides/asset-modules/

src/main/webapp/app/shared/badge/count-badge.scss
zhx828: @calvinlu3 there is no easy way I imagine because they from an image. Ideally, we should create our own font lib to use this icons. We can leave them as it for now.

src/main/webapp/app/shared/badge/CountBadge.tsx
```diff
@@ -0,0 +1,30 @@
...
+  max?: number;
+}
+
+const CountBadge = (props: CountBadgeProps) => {
+  const count = props.count > (props.max || Number.MAX_SAFE_INTEGER) ? `${props.max}+` : props.count;
+  const hideWhenOne = props.hideWhenOne !== undefined ? props.hideWhenOne : false;
+
+  return (
+    <div className="mr-1">
+      <div className={classNames('pt-2', hideWhenOne ? 'pr-0' : 'pr-2', 'count-badge-wrapper')}>
```
zhx828, line 18: pr-0 looks a bit off when other icons within the summary do have pr-2. Could it be only at pr-0 if no other badge uses pr-2?

zhx828 (changes requested):
One minor request: for mutation that is not string(with comma in the name), can we always show oncogenicity icon. similar to cbioportal?

Otherwise, looks good! Thanks.

zhx828 (approved):
LGTM, thanks!

PR #142, opened 2023-12-05
Add time series

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3622

### DIFF
--- a/src/main/webapp/app/components/timeSeries/TimeSeries.tsx
+++ b/src/main/webapp/app/components/timeSeries/TimeSeries.tsx
@@ -1,0 +1,185 @@
+import { APP_HISTORY_FORMAT, APP_TIME_FORMAT } from 'app/config/constants/constants';
+import React, { useEffect, useRef, useState } from 'react';
+import { Timeline, TimelineEvent } from 'react-event-timeline';
+import { TextFormat } from 'react-jhipster';
+import { Col, Row } from 'reactstrap';
+import './time-series.scss';
+import { groupTimeSeriesDataByDay } from './time-series-utils';
+
+export type RequiredTimeSeriesEventData = {
+  createdAt: Date;
+  editBy: string;
+  operation: string;
+  bubbleColor: string;
+  content: React.ReactNode;
+};
+
+export type ExtraTimeSeriesEventData = RequiredTimeSeriesEventData & {
+  title: string;
+  subtitle: string;
+};
+interface ITimeSeriesPropsGroupByDay {
+  groupByDay?: true;
+  data: ExtraTimeSeriesEventData[];
+}
+
+interface ITimeSeriesPropsNoGroupByDay {
+  groupByDay: false;
+  data: RequiredTimeSeriesEventData[];
+}
+
+export type ITimeSeriesProps = ITimeSeriesPropsGroupByDay | ITimeSeriesPropsNoGroupByDay;
+
+const TimeSeries = ({ data, groupByDay = true }: ITimeSeriesProps) => {
+  const dataForEachDay = groupTimeSeriesDataByDay(data);
+
+  const days = Object.keys(dataForEachDay);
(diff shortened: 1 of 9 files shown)


src/main/webapp/app/components/timeSeries/TimeSeries.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+export type RequiredTimeSeriesEventData = {
+  createdAt: Date;
+  editBy: string;
+  operation: string;
+  bubbleColor: string;
+  content: React.ReactNode;
+};
+
+export type ExtraTimeSeriesEventData = RequiredTimeSeriesEventData & {
```
zhx828, line 17: Why not merge these two into `RequiredTimeSeriesEventData`?

src/main/webapp/app/components/timeSeries/time-series-utils.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    dataForEachDay[dateString].push(eventData);
+  }
+
+  for (const eventData of Object.values<RequiredTimeSeriesEventData[] | ExtraTimeSeriesEventData[]>(dataForEachDay)) {
+    eventData.sort((event1, event2) => event2.createdAt.getTime() - event1.createdAt.getTime());
+  }
+
+  const days = Object.keys(dataForEachDay);
+  days.sort((day1, day2) => new Date(day2).getTime() - new Date(day1).getTime());
```
zhx828, line 22: Is this needed?

src/main/webapp/app/components/timeSeries/TimeSeries.tsx
```diff
@@ -0,0 +1,186 @@
...
+  createdAt,
+  editBy,
+  operation,
+  bubbleColor,
+  content,
+  groupByDay = true,
+}: ITimeSeriesEventProps) => {
+  return (
+    <TimelineEvent
+      style={{ width: '100%' }}
```
zhx828, line 147: Do we need to specify to make it work 100%?

zhx828 (commented):
Few minor comments, overall looks good.

src/main/webapp/app/components/timeSeries/TimeSeries.tsx
bprize15: I need the two to be separate so I can say if `groupByDay` is `true` or `undefined` (since it defaults to true), expect `ExtraTimeSeriesEventData` which includes `title` and subtitle`. Otherwise, just using `RequiredTimeSeriesData` which does not include those fields (since the non-grouped view does not use the title and subtitle)

src/main/webapp/app/components/timeSeries/time-series-utils.ts
bprize15: After getting grouping the timestamps by day, I know that the timestamps within each day are sorted, but there is no guarantee the days themselves are sorted, so I need to sort the days to display in the correct order

src/main/webapp/app/components/timeSeries/TimeSeries.tsx
bprize15: You're right, I do not. Will update

src/main/webapp/app/components/timeSeries/time-series-utils.ts
zhx828: I'm a bit surprised that if you sort days, the dataForEachDay is sorted as well. I think thats what you trying to do here?

bprize15: I do two sorts. The logic to construct the data structure goes as follows:

1. Create an unsorted mapping of day to all timestamps within that day.
2. For each array of timestamps sort. the timestamps by time.
3. Sort the keys from the map so the days are in order,

PR #144, opened 2023-12-08
Add gene timeline

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3622

### DIFF
--- a/src/main/webapp/app/components/geneHistoryTooltip/GeneHistoryTooltip.tsx
+++ b/src/main/webapp/app/components/geneHistoryTooltip/GeneHistoryTooltip.tsx
@@ -1,0 +1,70 @@
+import React, { useMemo } from 'react';
+import TimeSeries, { ExtraTimeSeriesEventData, RequiredTimeSeriesEventData } from '../timeSeries/TimeSeries';
+import './gene-history.scss';
+import { CANCER_TYPE_THERAPY_INDENTIFIER } from 'app/config/constants/constants';
+import DefaultTooltip from 'app/shared/tooltip/DefaultTooltip';
+import { FaHistory } from 'react-icons/fa';
+import { ParsedHistoryRecord } from 'app/pages/curation/CurationPage';
+import constructTimeSeriesData from './gene-history-tooltip-utils';
+
+export interface IGeneHistoryTooltipProps {
+  historyData: Map<string, ParsedHistoryRecord[]>;
+  location?: string;
+  groupByDay?: boolean;
+  contentFieldWhenObject?: string;
+}
+
+function GeneHistoryTooltip({ location, historyData, contentFieldWhenObject = 'description' }: IGeneHistoryTooltipProps) {
+  const parsedHistoryData = useMemo(() => {
+    if (!historyData) {
+      return [];
+    }
+
+    const parsedData: (RequiredTimeSeriesEventData | ExtraTimeSeriesEventData)[] = [];
+    for (const [recordLocation, records] of historyData) {
+      if (!location || recordLocation === location) {
+        addTimeSeriesDataForParsedRecords(records, parsedData);
+      } else if (location.startsWith(CANCER_TYPE_THERAPY_INDENTIFIER)) {
+        const parsedLocation = location.slice(CANCER_TYPE_THERAPY_INDENTIFIER.length);
+
+        const recordLocationSubstrings = recordLocation.split(',');
+        const parsedRecordLocationSubstrings = [];
+        for (const substr of recordLocationSubstrings) {
+          if (!substr.includes('_THERAPEUTIC_IMPLICATIONS_FOR_DRUG_') && !substr.includes('implications for')) {
+            parsedRecordLocationSubstrings.push(substr);
+          }
+        }
+        const parsedRecordLocation = parsedRecordLocationSubstrings.join(',');
+
+        if (parsedRecordLocation === parsedLocation) {
+          addTimeSeriesDataForParsedRecords(records, parsedData);
+        }
+      }
+    }
+    return parsedData;
+  }, [historyData, location]);
+
+  function addTimeSeriesDataForParsedRecords(
+    parsedRecords: ParsedHistoryRecord[],
+    parsedData: (RequiredTimeSeriesEventData | ExtraTimeSeriesEventData)[]
+  ) {
+    for (const parsedRecord of parsedRecords) {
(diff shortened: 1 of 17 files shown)


zhx828:
For the history at the bottom:

- I tried the diff section without any background(white), with our default border color (light grey). I think I like that better.
- For added section, can we only show first 500 letters? I remember you have the 300 as the default for the last comment. Maybe we could default for all TextEllipsis (size=sm, show 300, size=lg, show 500) We can also folded like what you current have if that's too much of work.
- We don't need to bold the user's name
- Use title/subtitle style, show location(gene/mutation/cancer type) as title, and which  section as subtitle. Title has mt-2, mb-1
- the diff section should take the whole with, instead of 98%
- let's use orange/warning color for name change/update

### current
Screenshot 2023-12-08 at 2 15 24 PM

### suggested UI
Screenshot 2023-12-08 at 2 15 30 PM


For the history in the tooltip:

### current
Screenshot 2023-12-08 at 2 38 54 PM

### suggested UI
Screenshot 2023-12-08 at 2 29 45 PM

zhx828 (changes requested):
Need @calvinlu3 's approval as well.

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -237,83 +316,110 @@ const CurationPage = (props: ICurationPageProps) => {
...
+                                          options={[]}
+                                        />
+                                        <RealtimeTextAreaInput
+                                          fieldKey={`mutations/${mutationIndex}/tumors/${tumorIndex}/TIs/${tiIndex}/treatments/${treatmentIndex}/description`}
+                                          inputClass={styles.textarea}
+                                          label="Description of Evidence"
+                                          labelIcon={
+                                            <GeneHistoryTooltip
+                                              historyData={parsedHistoryList}
+                                              location={`${CANCER_TYPE_THERAPY_INDENTIFIER}${getMutationName(
```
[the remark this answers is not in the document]
bprize15: The tooltip looks for that identifier since the location for cancer type has so many variations. When the tooltip sees this identifier it knows to do some special parsing for the location. I could have just checked explicitly for the character represented by that constant (&), but if the identifier passed into the tooltip were changed, the logic for parsing it would need to reflect that change in the tooltip. By making the identifier a constant, any changes to the identifier will not require changes to the parsing logic.

src/main/webapp/app/shared/util/firebase/firebase-utils.ts
```diff
@@ -68,6 +68,21 @@ export const getTxName = (drugList: DrugCollection, txUuidName: string) => {
...
 
+export const getTxId = (drugList: DrugCollection, txUuidName: string) => {
+  return txUuidName
+    .split(',')
+    .map(tx => {
+      return tx
+        .split('+')
+        .map(drug => {
+          drug = drug.trim();
+          return drugList[drug] ? drugList[drug].uuid : '';
```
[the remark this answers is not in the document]
bprize15: You're right. I didn't realize `drug` is `uuid`. I just took logic from a similar method.

src/main/webapp/app/components/geneHistoryTooltip/GeneHistoryTooltip.tsx
```diff
@@ -0,0 +1,174 @@
...
+    } catch {
+      body = <h6 className="gene-history-event-content gene-history-error-content">Record formatted incorrectly</h6>;
+    }
+
+    return <div>{body}</div>;
+  }
+
+  return (
+    <DefaultTooltip
+      disabled={parsedHistoryData.length === 0}
```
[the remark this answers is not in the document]
bprize15: Sure, I'll leave the color as grey to indicate to indicate `No history available` before hovering. But it won't actually be disabled.

zhx828 (approved):
Thanks.

src/main/webapp/app/shared/firebase/input/RealtimeBasicInput.tsx
```diff
@@ -14,16 +14,20 @@ export interface IRealtimeBasicLabel extends LabelProps {
...
   label,
   labelClass,
+  labelIcon,
   id,
   ...labelProps
 }: IRealtimeBasicLabel) => {
   const labelComponent = (
     <Label {...labelProps} id={id} for={id} className={classnames(labelClass, 'text-nowrap')}>
       {label}
+      {labelIcon && <span style={{ marginRight: '8px' }} />}
```
calvinlu3, line 29: Use className `mr-2` here?

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -192,6 +265,12 @@ const CurationPage = (props: ICurationPageProps) => {
                           fieldKey={`mutations/${mutationIndex}/mutation_effect/description`}
                           inputClass={styles.textarea}
                           label="Description of Evidence"
+                          labelIcon={
```
calvinlu3, line 268: There will be a history icon next to each mutation effect radio button label?

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -171,7 +233,18 @@ const CurationPage = (props: ICurationPageProps) => {
                           }))}
                         />
                         <RealtimeCheckedInputGroup
-                          groupHeader="Mutation Effect"
+                          groupHeader={
+                            <>
+                              <span style={{ marginRight: '8px' }}>Mutation Effect</span>
```
calvinlu3, line 238: We can use bootstrap margin class in the places where we use `8px`. Or is there some alignment requirements where it must be 8px?

src/main/webapp/app/pages/curation/CurationPage.tsx
bprize15: Yeah, we should have history for all of them (in theory)

src/main/webapp/app/pages/curation/CurationPage.tsx
bprize15: Nope, switched to mr-2

src/main/webapp/app/pages/curation/CurationPage.tsx
calvinlu3: Nevermind, the Github diff viewer made it look like you added groupHeader and labelIcon Props. Makes sense now

calvinlu3 (approved):

PR #145, opened 2023-12-11
Add no entry badge to Collapsible header

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3625

image

image

### DIFF
--- a/src/main/webapp/app/shared/util/firebase/firebase-utils.ts
+++ b/src/main/webapp/app/shared/util/firebase/firebase-utils.ts
@@ -104,3 +104,45 @@ export const isSectionRemovableWithoutReview = (geneData: Gene, nestLevel: Remov
   const review: Review = getValueByNestedKey(geneData, `${pathDetails.pathFromGene}/${reviewKey}`);
   return !!review && !!review.added;
 };
+
+export function isNestedObjectEmpty(obj: any, ignoredKeySubstrings: string[] = []) {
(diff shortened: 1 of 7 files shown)


src/main/webapp/app/shared/util/firebase/firebase-utils.ts
```diff
@@ -104,3 +104,62 @@ export const isSectionRemovableWithoutReview = (geneData: Gene, nestLevel: Remov
   const review: Review = getValueByNestedKey(geneData, `${pathDetails.pathFromGene}/${reviewKey}`);
   return !!review && !!review.added;
 };
+
+export function isNestedObjectEmpty(obj: any, ignoredKeySubstrings: string[] = []) {
+  if (typeof obj === 'object' && obj !== undefined && obj !== null) {
+    let targetKeys = Object.keys(obj);
+    if (ignoredKeySubstrings !== undefined && ignoredKeySubstrings.length > 0) {
+      targetKeys = Object.keys(obj).filter(key => !ignoredKeySubstrings.some(suffix => key.includes(suffix)));
```
zhx828, line 112: why to use includes, not ignorecaseEqual?

calvinlu3: I'm trying to filter out the keys that includes `_review or _uuid`. Keys like `name_review, tsg_review, ...` should not be taken into account when determining if a section is empty, so need to check for substring using `includes()`.

PR #146, opened 2023-12-12
Add sidebar

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3623


src/main/webapp/app/components/sidebar/GenericSidebar.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <div style={{ marginTop: '2rem', marginLeft: '1rem', display: 'flex' }}>
+        <FaChevronRight cursor={'pointer'} style={{ marginTop: '6px' }} onClick={() => toggleGenericSidebar(false)} />
+        {children}
+      </div>
+    </div>
+  ) : (
+    <div className="generic-sidebar-collapsed">
+      <FaChevronLeft
+        cursor={'pointer'}
+        style={{ position: 'absolute', right: '38', top: '38' }}
```
zhx828, line 24: What's this magic 38 number?

src/main/webapp/app/components/sidebar/GenericSidebar.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { observer } from 'mobx-react';
+import { IRootStore } from 'app/stores';
+import { componentInject } from 'app/shared/util/typed-inject';
+import { FaChevronRight, FaChevronLeft } from 'react-icons/fa';
+
+export interface IGenericSidebarProps extends StoreProps {
+  children: React.ReactNode;
+}
+
+const GenericSidebar = ({ children, showGenericSidebar, toggleGenericSidebar, genericSidebarWidth }: IGenericSidebarProps) => {
```
zhx828, line 12: Should we just call is Sidebar? Will it collide with components from other library?

zhx828 (changes requested):

src/main/webapp/app/components/sidebar/GenericSidebar.tsx
bprize15: It's just a position I chose that makes the chevron that opens the sidebar look better. It also aligns it with the y position of the chevron that closes the sidebar, so the transition is smooth. Should we make it a constant?

src/main/webapp/app/components/sidebar/GenericSidebar.tsx
bprize15: It will collide. Changing it to OncoKBSidebar as discussed yesterday

zhx828 (approved):
LGTM, thanks.

PR #149, opened 2023-12-22
Updates

zhx828:
- Replace buttons in entity detail page && Minor fixes

- Add association importer

- Update LoE table to include level description/html description/color


calvinlu3 (approved):
LGTM

PR #150, opened 2023-12-26
Add sidebar in curation page and display history

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3624

### DIFF
--- a/src/main/webapp/app/components/geneHistoryTooltip/gene-history-tooltip-utils.tsx
+++ b/src/main/webapp/app/components/geneHistoryTooltip/gene-history-tooltip-utils.tsx
@@ -1,7 +1,8 @@
-import { HistoryRecord, HistoryRecordState } from 'app/shared/model/firebase/firebase.model';
+import { DrugCollection, HistoryRecord, HistoryRecordState } from 'app/shared/model/firebase/firebase.model';
 import { RequiredTimeSeriesEventData, ExtraTimeSeriesEventData } from '../timeSeries/TimeSeries';
 import ReactDiffViewer from 'react-diff-viewer-continued';
 import React from 'react';
+import { getTxName } from 'app/shared/util/firebase/firebase-utils';
 
 export default function constructTimeSeriesData(
   record: HistoryRecord,
@@ -50,6 +51,8 @@ export default function constructTimeSeriesData(
     operation,
     bubbleColor,
     content,
+    location: record.location,
+    objectField,
   };
 }
 
@@ -62,7 +65,7 @@ export function getTimeSeriesDataContent(objectField: string, newContent: Histor
         body = (
           <div>
             <div className="gene-history-event-content">
-              <ReactDiffViewer oldValue={oldContent} newValue={newContent} splitView={false} hideLineNumbers />
+              <ReactDiffViewer showDiffOnly={false} oldValue={oldContent} newValue={newContent} splitView={false} hideLineNumbers />
             </div>
           </div>
         );
@@ -75,7 +78,13 @@ export function getTimeSeriesDataContent(objectField: string, newContent: Histor
         body = (
           <div>
             <div className="gene-history-event-content">
-              <ReactDiffViewer oldValue={oldContent?.[objectField]} newValue={newContent[objectField]} splitView={false} hideLineNumbers />
+              <ReactDiffViewer
+                showDiffOnly={false}
+                oldValue={oldContent?.[objectField]}
+                newValue={newContent[objectField]}
+                splitView={false}
+                hideLineNumbers
+              />
             </div>
           </div>
         );
@@ -88,4 +97,35 @@ export function getTimeSeriesDataContent(objectField: string, newContent: Histor
   }
 
   return <div>{body}</div>;
+}
+
+export function formatLocation(location: string, drugList: DrugCollection, objectField: string) {
+  const locationSubstrings = location.split(',');
+  const lastSubstring = locationSubstrings[locationSubstrings.length - 1];
+  if (lastSubstring.endsWith('Mutation Effect')) {
+    if (objectField === 'description') {
+      return location + ', Description of Evidence';
+    } else if (objectField === 'oncogenic') {
+      return location + ', Oncogenic';
+    }
+  } else if (lastSubstring.endsWith('Tumor Type Summary')) {
+    return locationSubstrings.slice(0, -1).join(',') + ', Therapeutic Summary';
+  } else {
+    let index = -1;
+    for (let i = 0; i < locationSubstrings.length; i++) {
+      if (
+        locationSubstrings[i].includes('_THERAPEUTIC_IMPLICATIONS_FOR_DRUG_') ||
+        locationSubstrings[i].includes('implications for') ||
+        locationSubstrings[i].includes('INVESTIGATIONAL_THERAPEUTIC_IMPLICATIONS_DRUG')
+      ) {
+        index = i;
+        break;
+      }
+    }
+
+    if (index > -1) {
+      return locationSubstrings.slice(0, index) + ', ' + getTxName(drugList, locationSubstrings.slice(index + 1).join(','));
+    }
+  }
+  return location;
 }
--- a/src/main/webapp/app/components/tabs/tabs.tsx
+++ b/src/main/webapp/app/components/tabs/tabs.tsx
@@ -1,0 +1,47 @@
+import React, { useEffect, useState } from 'react';
+import { Col, Container, Row } from 'reactstrap';
+import { observer } from 'mobx-react-lite';
+import classNames from 'classnames';
+import { IRootStore } from 'app/stores/createStore';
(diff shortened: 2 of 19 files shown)


zhx828:
Can you make some slight changes?
- the Author dropdown seems having both name and email, can we use the name only? 
- show `Clear Filters` button only when there is filter.
- We should display the last 10 changes in descending order
- Add some language at the bottom of the `Change history` section indicating we are only displaying the last 10 changes.
- The history tooltip is too narrow when there are updates. 
- I'm wondering whether we can remove the filter title when it's clearly for filtering, we could so something like below
Screenshot 2023-12-26 at 11 48 20 AM

src/main/webapp/app/components/Tabs/CurationHistoryTab.tsx
```diff
@@ -0,0 +1,259 @@
...
+          options.push({ value: email, label: data.name || `[${email}]` });
+        }
+      }
+      return options;
+    }, [usersData]);
+
+    useEffect(() => {
+      const cleanupCallbacks = [];
+      cleanupCallbacks.push(addUsersListener(firebaseUsersPath));
+      cleanupCallbacks.push(addDrugListListener());
```
zhx828, line 102: Why do we need these two?

src/main/webapp/app/stores/history-tab.store.tsx
```diff
@@ -0,0 +1,60 @@
+import { action, computed, makeObservable, observable } from 'mobx';
+import _ from 'lodash';
+
+export class HistoryTabStore {
```
zhx828, line 4: I don't think we gonna share the history tab store and it should be placed with the tab component together.

src/main/webapp/app/stores/tabs.store.ts
```diff
@@ -0,0 +1,16 @@
+import { action, makeObservable, observable } from 'mobx';
+
+export class TabsStore {
```
zhx828, line 3: The same to this store. The stores folder should only include the ones that intended to be shared across different pages. If it's a store that manages the state for a specific component, it needs to be placed together with component.

src/main/webapp/app/components/geneHistoryTooltip/gene-history-tooltip-utils.tsx
```diff
@@ -89,3 +98,34 @@ export function getTimeSeriesDataContent(objectField: string, newContent: Histor
 
   return <div>{body}</div>;
 }
+
+export function formatLocation(location: string, drugList: DrugCollection, objectField: string) {
```
zhx828, line 102: This method should be tested.

src/main/webapp/app/components/Tabs/tabs.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../variables.scss';
+
+.tab {
```
zhx828, line 3: Try to use module.scss instead for component with a very generic name, so the compiler will create a unique classname to prevent styling pollution

src/main/webapp/app/components/Tabs/tabs.tsx
```diff
@@ -0,0 +1,58 @@
...
+export type Tab = {
+  title: string;
+  content: React.ReactNode;
+};
+
+export interface ITabsProps extends StoreProps {
+  tabs: Tab[];
+}
+
+const Tabs = observer(({ tabs, tabsStore }: ITabsProps) => {
```
zhx828, line 18: Why do we create our own Tabs component?

src/main/webapp/app/components/Tabs/CurationHistoryTab.tsx
```diff
@@ -0,0 +1,259 @@
...
+                <span className="mr-2" />
+                <TextFormat value={data.createdAt} type="date" format={APP_TIME_FORMAT} />
+              </span>
+              <DefaultTooltip
+                overlayStyle={{ maxWidth: 600 }}
+                overlayInnerStyle={{ maxHeight: '400px', overflow: 'auto', maxWidth: 'inherit' }}
+                overlay={isContentAvailable ? data.content : 'No content available'}
+              >
+                <FaHistory className={!isContentAvailable && 'grey-tooltip'} />
+              </DefaultTooltip>
```
zhx828, line 167: In TimeSeries, you have a default tooltip when history is not available. Should we consolidate?

zhx828 (changes requested):
- The component folder should be named as `tabs` instead of `Tabs`

src/main/webapp/app/components/Tabs/CurationHistoryTab.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+type HistoryTabData = {
+  record: HistoryRecord;
+  timeStamp: string;
+  admin: string;
+  location: string;
+  objectField?: string;
+};
+
+const CurationHistoryTab = observer(
```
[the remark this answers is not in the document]
bprize15: It listens to the historyTabStore, which is just a regular MobX store, so the component needs to be observable to be notified of changes

src/main/webapp/app/components/Tabs/CurationHistoryTab.tsx
```diff
@@ -0,0 +1,259 @@
...
+            } else {
+              data.push({
+                record,
+                timeStamp: history.timeStamp,
+                admin: history.admin,
+                location: record.location,
+              });
+            }
+          }
+        } catch {
```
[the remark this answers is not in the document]
bprize15: Sometimes `historyData.records` is an object instead of an array. In that case, I just ignore it and continue.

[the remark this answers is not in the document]
bprize15: I could also change it to `if (Symbol.iterator in history.records)`

src/main/webapp/app/components/Tabs/CurationHistoryTab.tsx
bprize15: The Users listener is to populate the author dropdown from the firebase users. The drugList is necessary to display the location correctly. Since the location comes with a list of drug uuids, I need the drug list to retrieve the drug name (this is all inside `formatLocation` and uses the `getTxName`) helper function

src/main/webapp/app/components/Tabs/tabs.tsx
bprize15: I didn't realize reactstrap had its own Tabs component until now. This one wasn't too hard to create and isn't a very large file. I can try refactoring to use reactstrap if you think that would be less bloat.

src/main/webapp/app/components/Tabs/CurationHistoryTab.tsx
bprize15: Since I'm not using the TimeSeries component in the sidebar (since I could not remove the line on the left side) or the GeneHistoryTooltip, the only overlap between the two types of history we have `GeneHistoryTooltip` and `CurationHistoryTab` are their use of the utility functions. To do this, I would need to refactor all of this, so I think it is best to leave for now and refactor in another PR.

src/main/webapp/app/components/Tabs/CurationHistoryTab.tsx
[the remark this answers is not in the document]
zhx828: Would be better if we just handle the case directly so it has some contexts here.

src/main/webapp/app/components/Tabs/tabs.tsx
zhx828: I will not create component if we can use existing library unless we have special needs.

src/main/webapp/app/components/tabs/tabs.tsx
```diff
@@ -0,0 +1,57 @@
...
+export type Tab = {
+  title: string;
+  content: React.ReactNode;
+};
+
+export interface ITabsProps extends StoreProps {
+  tabs: Tab[];
+}
+
+const Tabs = observer(({ tabs, tabsStore }: ITabsProps) => {
```
zhx828, line 18: Can we make this a function component without the store, use react hooks instead? I don't see the benefits creating the store.

src/main/webapp/app/components/tabs/tabs.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const openTab = tabsStore.openTab;
+
+  useEffect(() => {
+    tabsStore.setOpenTab(tabs[0]?.title);
+  }, []);
+
+  return (
+    <Container style={{ wordBreak: 'break-word', marginRight: '16px', marginTop: '3.5px', marginLeft: '10px' }}>
```
zhx828, line 26: For margins/paddings, let's stick with the classnames `p-1`, `m-1` etc.

zhx828 (changes requested):
Let's move tabs.tsx under the shared folder. 
Feel free to merge after fixing my comments.

PR #151, opened 2023-12-27
add additional mutation filtering

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3626


src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -194,253 +328,349 @@ const CurationPage = (props: ICurationPageProps) => {
...
+                      </Collapsible>
+                      {mutation.mutation_effect.germline && (
                         <Collapsible
-                          nestLevel={NestLevelType.SOMATIC}
-                          title={'Somatic'}
-                          geneFieldKey={`mutations/${mutationIndex}/mutation_effect/oncogenic`}
+                          nestLevel={NestLevelType.GERMLINE}
+                          className={'mt-2'}
+                          title={'Germline'}
+                          geneFieldKey="mutations/0/mutation_effect/germline"
```
calvinlu3, line 430: Index shouldn't be `0` here

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -194,253 +328,349 @@ const CurationPage = (props: ICurationPageProps) => {
...
+                      color="danger"
+                      onClick={() => {
+                        setOncogenicityFilter(initFilterCheckboxState(oncogenicityFilterOptions));
+                        setMutationEffectFilter(initFilterCheckboxState(mutationEffectFilterOptions));
+                        setTxLevelFilter(initFilterCheckboxState(txLevelFilterOptions));
+                        setShowFilterModal(false);
+                      }}
+                    >
+                      Reset
+                    </Button>
```
calvinlu3, line 646: Wonder if we should also have a reset button next to the text `Showing x of y matching search`.

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -194,253 +328,349 @@ const CurationPage = (props: ICurationPageProps) => {
...
+                )}
+                <Col className="px-0 mr-2" style={{ flexGrow: 0 }}>
+                  <Button
+                    color="primary"
+                    onClick={() => {
+                      handleFilterMutations();
+                      setShowFilterModal(false);
+                    }}
+                  >
+                    Confirm
```
calvinlu3, line 657: Worth to discuss if we want to just have the filter results change as we click on checkboxes or make selections instead of having to make an extra click to apply filters.

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -194,253 +328,349 @@ const CurationPage = (props: ICurationPageProps) => {
...
+                {oncogenicityFilter.map((filter, index) => (
+                  <Col className="col-6" key={filter.label}>
+                    <InputGroup>
+                      <Input
+                        onChange={() => handleFilterCheckboxChange(index, setOncogenicityFilter)}
+                        checked={filter.selected}
+                        style={{ marginLeft: '0px' }}
+                        type="checkbox"
+                      />
+                      <Label style={{ marginLeft: '18.75px' }}>{filter.label}</Label>
```
calvinlu3, line 595: Can we add the `for` prop and also would need to add and `id` to the Input. I believe it supposed to make it so that when you click on the label, it also checks the box/radio

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -194,253 +328,349 @@ const CurationPage = (props: ICurationPageProps) => {
...
+                    </InputGroup>
+                  </Col>
+                ))}
+              </Row>
+              <h6 className="mb-2 mt-2">Therapeutic implication level of evidence</h6>
```
calvinlu3, line 616: Maybe could shorten to just Therapeutic Levels?

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -194,253 +328,349 @@ const CurationPage = (props: ICurationPageProps) => {
...
+                {txLevelFilter.map((filter, index) => (
+                  <Col style={{ flexGrow: 0 }} key={filter.label}>
+                    <InputGroup>
+                      <Input
+                        onChange={() => handleFilterCheckboxChange(index, setTxLevelFilter)}
+                        checked={filter.selected}
+                        style={{ marginLeft: '0px' }}
+                        type="checkbox"
+                      />
+                      <Label style={{ marginLeft: '18.75px' }}>{filter.label}</Label>
```
calvinlu3, line 627: I haven't looked at this PR locally, so wondering what the 18.75px is used for? Is it for centering the label?

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -194,253 +328,349 @@ const CurationPage = (props: ICurationPageProps) => {
...
+                    </InputGroup>
+                  </Col>
+                ))}
+              </Row>
+              <Row className="justify-content-end">
+                {isFiltered() && (
+                  <Col className="px-0 mr-2" style={{ flexGrow: 0 }}>
+                    <Button
+                      outline
+                      color="danger"
```
calvinlu3, line 637: We can probably just use the `btn-outline-primary`, but I see that it also conflicts with the save button. If we just show filtered results immediately, then it would be fine.

calvinlu3:
@bprize15 Just a couple of comments. Would be great to also get Hongxin to review once you've made the changes. Thanks!

src/main/webapp/app/pages/curation/CurationPage.tsx
bprize15: The checkbox from reactstrap has a left margin of -18.75px for some reason. This is to account for that and make the spacing even on both sides.

src/main/webapp/app/pages/curation/CurationPage.tsx
bprize15: Discussed with Hongxin. We changed it to a modal, so we the confirm button makes the most sense in this case.

zhx828:
- [ ] `Showing ...`  language should only be shown when there is a filter
- [ ] we should differentiate the filter icon when there is/not filter applied (similar to comment, maybe change to yellow when there is a filter)?
- [ ] Add `cursor: pointer;` on the filter icon
- [ ] Are you calculating the stats now? For instance, in ABL1, when I apply LoF, I see no mutation matched. Shouldn't that checkbox be disabled?

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -45,6 +47,65 @@ const CurationPage = (props: ICurationPageProps) => {
...
+  const [txLevelFilter, setTxLevelFilter] = useState(initFilterCheckboxState(TX_LEVEL_OPTIONS));
+  const [tempTxLevelFilter, setTempTxLevelFilter] = useState(initFilterCheckboxState(TX_LEVEL_OPTIONS));
+
+  const [enabledCheckboxes, setEnabledCheckboxes] = useState<string[]>(null);
+
+  function initFilterCheckboxState(options: string[]) {
+    return options.map(option => ({ label: option, selected: false, disabled: false }));
+  }
+
+  function mutationsAreFiltered() {
```
zhx828, line 68: This should be a computed value.

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -45,6 +47,65 @@ const CurationPage = (props: ICurationPageProps) => {
...
+    return (
+      oncogenicityFilter.some(filter => filter.selected) ||
+      mutationEffectFilter.some(filter => filter.selected) ||
+      txLevelFilter.some(filter => filter.selected)
+    );
+  }
+
+  function showFilterModalCancelButton() {
```
zhx828, line 76: This should be computed.

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -194,253 +332,349 @@ const CurationPage = (props: ICurationPageProps) => {
...
+                    <InputGroup>
+                      <Input
+                        id={`oncogenicity-filter-${filter.label}`}
+                        onChange={() => handleFilterCheckboxChange(index, setTempOncogenicityFilter)}
+                        checked={filter.selected}
+                        disabled={!enabledCheckboxes.includes(filter.label)}
+                        style={{ marginLeft: '0px' }}
+                        type="checkbox"
+                      />
+                      <Label for={`oncogenicity-filter-${filter.label}`} style={{ marginLeft: '18.75px' }}>
```
zhx828, line 585: Use class name instead of style?

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -194,253 +332,349 @@ const CurationPage = (props: ICurationPageProps) => {
...
+              <Row>
+                {tempOncogenicityFilter.map((filter, index) => (
+                  <Col className="col-6" key={filter.label}>
+                    <InputGroup>
+                      <Input
+                        id={`oncogenicity-filter-${filter.label}`}
+                        onChange={() => handleFilterCheckboxChange(index, setTempOncogenicityFilter)}
+                        checked={filter.selected}
+                        disabled={!enabledCheckboxes.includes(filter.label)}
+                        style={{ marginLeft: '0px' }}
```
zhx828, line 582: No need to change, but you can use `0`

src/main/webapp/app/pages/curation/CurationPage.tsx
zhx828: Agree with Calvin

zhx828 (changes requested):

src/main/webapp/app/pages/curation/CurationPage.tsx
bprize15: I wanted an exact left margin of 18.75 pixels. The checkbox originally comes with a left margin of -18.75px which I needed to set to 0 to fix the spacing. I use the above to offset this. I tried using `ml-3` and `ml-4`, but they both looked off to me. Let me know what you think:
Screenshot 2024-01-02 at 4 54 45 PM
Screenshot 2024-01-02 at 4 54 57 PM

zhx828: Makes sense. Then could you make it a constant and add a comment?

zhx828 (approved):
LGTM, thanks! Once the conflict is resolved, feel free to merge.

PR #152, opened 2023-12-27
Fix firebase meta collaborators list deletion

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3641


bprize15 (commented):
LGTM

PR #153, opened 2023-12-29
Update evidence model & Add firebase models

zhx828:
Final PR before deploying 🙏🏼 
I will send data for testing.


src/main/java/org/mskcc/oncokb/curation/service/FirebaseService.java
```diff
@@ -0,0 +1,93 @@
...
+        AssociationService associationService
+    ) {
+        this.geneService = geneService;
+        this.alterationService = alterationService;
+        this.evidenceService = evidenceService;
+        this.associationService = associationService;
+    }
+
+    public void readGene() {
+        String PATH = "Genes/BRAF";
```
calvinlu3, line 43: I see that FirebaseService is not used at the moment, but its supposed to get the gene data from FB and convert to FB Java models so that we can use it based on my understanding.

Should the gene name be a param that is passed?

src/main/java/org/mskcc/oncokb/curation/domain/firebase/Gene.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class Gene {
+
+    String name;
+    String summary;
+    String background;
+    GeneType type;
+    List<Mutation> mutations = new ArrayList<>();
```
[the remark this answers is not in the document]
zhx828: This is just for demo purpose. Not a complete list, I will send a separate PR to finish the importing work.

src/main/java/org/mskcc/oncokb/curation/service/FirebaseService.java
zhx828: This is just for demo purpose. Not a complete list, I will send a separate PR to finish the importing work.

PR #154, opened 2024-01-02
Minor changes to curation page

zhx828:
- Open mutation effect/somatic/germline when mutation is opened
- Lower minimum height for summaries
- Display diagnostic/prognostic implication and summary
- Show default option in RealtimeDropdownInput if available


calvinlu3 (approved):
Looks good to me

PR #155, opened 2024-01-02
Curator add gene

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3639


src/main/webapp/app/components/tabs/CurationToolsTab.tsx
```diff
@@ -0,0 +1,75 @@
...
+      setCreateButtonDisabled(true);
+    }
+  }, [selectedGene]);
+
+  async function handleCreateGene() {
+    const gene = new Gene(selectedGene);
+    await Promise.all([
+      createGene(getFirebasePath('GENE', gene.name), gene),
+      createMetaGene(getFirebasePath('META_GENE', gene.name), new Meta()),
+    ]);
```
zhx828, line 35: Need to handle the exception when one of the promises failed, we should roll back creation.

src/main/webapp/app/components/tabs/CurationToolsTab.tsx
```diff
@@ -0,0 +1,75 @@
...
+          <h6>Curate New Gene</h6>
+        </Col>
+      </Row>
+      <Row className="mb-2">
+        <Col className="px-0">
+          <GeneSelect onChange={option => setSelectedGene(option?.label)} />
+        </Col>
+      </Row>
+      <Row className={`align-items-center ${createButtonDisabled && selectedGene ? 'justify-content-between' : 'justify-content-end'}`}>
+        {createButtonDisabled && selectedGene && (
```
zhx828, line 52: The logic of when`createButtonDisabled && selectedGene`, then gene already exists should be a computed value. Or probably just rename `createButtonDisabled`?

zhx828 (changes requested):

src/main/webapp/app/components/tabs/CurationToolsTab.tsx
bprize15: I changed the component quite a bit. `selectedGene` is now a ref since it changing does not affect the UI (the GeneSelect handles its own rendering). 

`createButtonDisabled` and `showGeneExistsWarning` now are the only stateful variables. They both need to exist since if a gene is not selected, the button should be disabled, but the warning should not be shown.

zhx828 (approved):
Makes more sense with the latest implementation. Thanks!

PR #157, opened 2024-01-03
Add uuid to Drug entity & Add importer to update drug uuid & Add importer to verify gene hugo

zhx828:


src/main/java/org/mskcc/oncokb/curation/importer/CoreImporter.java
```diff
@@ -45,11 +48,15 @@ public CoreImporter(
         this.drugService = drugService;
         this.nciThesaurusService = nciThesaurusService;
         this.mainService = mainService;
+        this.applicationProperties = applicationProperties;
+
+        DATA_DIRECTORY = applicationProperties.getOncokbDataRepoPath() + "/curation/oncokb/";
```
calvinlu3, line 53: Not sure if we want to keep consistent, but we had a naming convention where DIRECTORY is the path without the slash and DIR_PATH is with slash

src/main/java/org/mskcc/oncokb/curation/service/CategoricalAlterationService.java
```diff
@@ -62,6 +66,13 @@ public Optional<CategoricalAlteration> partialUpdate(CategoricalAlteration categ
             .map(categoricalAlterationRepository::save);
     }
 
+    public Optional<CategoricalAlteration> findByAlterationTypeAndName(AlterationType alterationType, String name) {
+        if (alterationType == null) {
+            return Optional.empty();
+        }
+        return categoricalAlterationRepository.findByAlterationTypeAndName(alterationType, name);
```
calvinlu3, line 73: Shouldn't the spring jpa method already return empty optional if type is null?

zhx828: I want to avoid sending call when alteration type is null

calvinlu3 (approved):

PR #161, opened 2024-01-09
Modify cancer type modal

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/225

### DIFF
--- a/src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
+++ b/src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import Collapsible from './Collapsible';
 import { IRootStore } from 'app/stores';
 import { componentInject } from 'app/shared/util/typed-inject';
@@ -16,9 +16,13 @@ import { MUTATION_EFFECT_OPTIONS, ONCOGENICITY_OPTIONS } from 'app/config/consta
 import styles from './styles.module.scss';
 import { AutoParseRefField } from 'app/shared/form/AutoParseRefField';
 import { CANCER_TYPE_THERAPY_INDENTIFIER, GERMLINE_INHERITANCE_MECHANISM, PATHOGENICITY, PENETRANCE } from 'app/config/constants/constants';
-import { getCancerTypeName } from 'app/shared/util/utils';
+import { generateUuid, getCancerTypeName } from 'app/shared/util/utils';
 import CancerTypeLevelSummary from '../nestLevelSummary/CancerTypeLevelSummary';
 import RealtimeDropdownInput from 'app/shared/firebase/input/RealtimeDropdownInput';
+import { FaEdit } from 'react-icons/fa';
+import ModifyCancerTypeModal from 'app/shared/modal/ModifyCancerTypeModal';
+import _ from 'lodash';
+import { notifyError } from 'app/oncokb-commons/components/util/NotificationUtils';
 import TreatmentLevelSummary from '../nestLevelSummary/TreatmentLevelSummary';
 
 export interface IMutationCollapsibleProps extends StoreProps {
@@ -31,6 +35,7 @@ const MutationCollapsible = ({
   data,
   hugoSymbol,
   deleteSection,
+  updateTumor,
   drugList,
   mutation,
   firebaseIndex,
@@ -39,6 +44,8 @@ const MutationCollapsible = ({
   const title = getMutationName(mutation);
   const mutationFirebasePath = buildFirebaseGenePath(hugoSymbol, `mutations/${firebaseIndex}`);
   const showMutationLevelSummary = !title.includes(',');
+
+  const [openCancerTypesUuid, setOpenCancerTypesUuid] = useState(null);
 
   return (
     <Collapsible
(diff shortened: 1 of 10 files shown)


zhx828 (changes requested):
When user delete all cancer types from the dropdown, the confirm button should be disabled. We should find a place to mention cancer type can not be empty, otherwise they should delete the cancer type section instead. Currently results to an error page.

src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
```diff
@@ -173,188 +179,216 @@ const MutationCollapsible = ({
...
+              excludedCancerTypes={tumor.excludedCancerTypes || []} // update model to include excluded cancer types
+              open={openCancerTypesUuid === tumor.cancerTypes_uuid}
+              onConfirm={async (includedCancerTypes, excludedCancerTypes) => {
+                const newTumor = _.cloneDeep(tumor);
+                newTumor.cancerTypes = includedCancerTypes;
+                newTumor.excludedCancerTypes = excludedCancerTypes;
+                if (!newTumor.excludedCancerTypes_uuid) {
+                  newTumor.excludedCancerTypes_uuid = generateUuid();
+                }
+                await updateTumor(cancerTypeFirebasePath, newTumor);
```
zhx828, line 383: @calvinlu3 what's our general strategy when firebase update fails for any reason?

src/main/webapp/app/shared/modal/ModifyCancerTypeModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+    const cancerType = new CancerType();
+    cancerType.code = cancerTypeSelectOption.code;
+    cancerType.mainType = cancerTypeSelectOption.mainType;
+    cancerType.subtype = cancerTypeSelectOption.subtype;
+    return cancerType;
+  }
+
+  async function getICancerTypeFromCancerType(cancerType: CancerType) {
+    try {
+      const iCancerType = await searchCancerTypes({ query: cancerType.subtype ? cancerType.subtype : cancerType.mainType });
```
zhx828, line 61: let's use code over subtype. The subtype can collide with mainType.

src/main/webapp/app/shared/modal/ModifyCancerTypeModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+      return iCancerType['data'][0];
+    } catch {
+      return null;
```
zhx828, line 64: I think this will be wrong if we return null. I would rather disable the functionality of modifying when error happens.

src/main/webapp/app/shared/modal/ModifyCancerTypeModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+          const option: CancerTypeSelectOption = {
+            label: getCancerTypeName(cancerType),
+            value: cancerType.id,
+            code: cancerType.code,
+            mainType: cancerType.mainType,
+            subtype: cancerType.subtype,
+          };
+          return [...accumulator, option];
+        }, [])
```
zhx828, line 88: This logic is dup of the one in getICancerTypeFromCancerType? Can we reuse?

src/main/webapp/app/shared/modal/ModifyCancerTypeModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+      onConfirm={() => {
+        onConfirm(
+          included.map(option => getCancerTypeFromCancerTypeSelectOption(option)),
+          excluded.map(option => getCancerTypeFromCancerTypeSelectOption(option))
+        );
+      }}
+      body={
+        <Container className="px-0">
+          <Row className="my-2">
+            <Col>
```
zhx828, line 139: Do we need to use container/row/col if you are not using the grids?

src/main/webapp/app/shared/modal/ModifyCancerTypeModal.tsx
```diff
@@ -0,0 +1,180 @@
...
+              <h6>Select cancer type(s) for INCLUSION</h6>
+            </Col>
+          </Row>
+          <Row>
+            <Col>
+              <CancerTypeSelect value={included} onChange={options => setIncluded(options)} isMulti />
+            </Col>
+          </Row>
+          {confirmDisabled ? (
```
zhx828, line 148: `confirmDisabled` should be changed to a different name here. The context doesn't add up.

zhx828 (changes requested):

src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
zhx828: Also our strategy overall on transection when there are multiple edits across different firebase paths within same transection.

src/main/webapp/app/shared/modal/ModifyCancerTypeModal.tsx
bprize15: Oh yeah, no need anymore since we removed horizontally aligned elements.

src/main/webapp/app/shared/modal/ModifyCancerTypeModal.tsx
bprize15: I disabled the dropdowns and the confirm button, added an error message, and a retry button to  retry fetching the data

zhx828 (approved):

PR #166, opened 2024-01-12
Modify therapy modal

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/228


src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
```diff
@@ -309,69 +312,96 @@ const MutationCollapsible = ({
...
+                              mutationUuid={mutation.name_uuid}
+                              cancerTypesUuid={tumor.cancerTypes_uuid}
+                              treatmentUuid={treatment.name_uuid}
                             />
                           }
-                          name="evidenceDescription"
+                          action={
+                            <>
+                              <FaEdit
+                                style={{ cursor: 'pointer' }}
```
zhx828, line 331: Is there a way we can make this style applicable to all FaEdit usage?

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -35,6 +35,7 @@ import CurationHistoryTab from 'app/components/tabs/CurationHistoryTab';
 import { FaFilter } from 'react-icons/fa';
 import _ from 'lodash';
 import MutationCollapsible from './collapsible/MutationCollapsible';
+import ModifyTherapyModal from 'app/shared/modal/ModifyTherapyModal';
```
zhx828, line 38: not used

src/main/webapp/app/shared/modal/modify-therapy-modal.store.tsx
```diff
@@ -0,0 +1,61 @@
+import { action, makeObservable, observable } from 'mobx';
+import { DrugSelectOption } from '../select/DrugSelect';
+
+export class ModifyTherapyModalStore {
```
zhx828, line 4: I didn't expect this to be a global store, but rather only exists when modal opens. It should be destroyed after closing.

zhx828 (changes requested):

src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
bprize15: We now have `OncoKBEdit`

src/main/webapp/app/shared/modal/modify-therapy-modal.store.tsx
bprize15: Will readdress after listener refactor

src/main/webapp/app/components/icons/OncoKBEdit.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { IconBaseProps } from 'react-icons';
+import { FaEdit } from 'react-icons/fa';
+import { IconContext } from 'react-icons/lib';
+
+export default function OncoKBEdit(props: IconBaseProps) {
```
zhx828, line 6: This should be moved to shared/icons, and should be named as EditIcon. Same goes to CommentIcon, it should be moved as well.

zhx828 (changes requested):

PR #170, opened 2024-01-18
Add option to search gene by exact hugo symbol

zhx828:
I haven't found a good way to query string value and order based on exact match/startsWith/contains. I tried the jpa query `blurSearchByHugoSymbol`. But it only supports one union currently.

### DIFF
--- a/src/main/java/org/mskcc/oncokb/curation/service/GeneService.java
+++ b/src/main/java/org/mskcc/oncokb/curation/service/GeneService.java
@@ -200,8 +200,8 @@ public class GeneService {
         return geneRepository.findByHugoSymbolInIgnoreCase(searchValues.stream().collect(Collectors.toList()));
     }
 
-    public Page<Gene> search(String query, Pageable pageable) {
-        Page<Gene> genePage = geneRepository.searchGene(query, pageable);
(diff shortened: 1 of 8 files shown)


src/main/java/org/mskcc/oncokb/curation/service/GeneService.java
```diff
@@ -200,8 +200,8 @@ public List<Gene> findByHugoSymbolIn(Set<String> searchValues) {
         return geneRepository.findByHugoSymbolInIgnoreCase(searchValues.stream().collect(Collectors.toList()));
     }
 
-    public Page<Gene> search(String query, Pageable pageable) {
-        Page<Gene> genePage = geneRepository.searchGene(query, pageable);
+    public Page<Gene> blurSearchByHugoSymbol(String query, Pageable pageable) {
```
calvinlu3, line 203: Where is this method use? Or do you plan on using it?

zhx828: Not used, just want to add an example on how to do blur search on repository level.

calvinlu3 (approved):
LGTM

PR #171, opened 2024-01-19
Add Release Gene to Curation Page Tools Tab

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3651

### DIFF
--- a/src/main/webapp/app/components/tabs/CurationToolsTab.tsx
+++ b/src/main/webapp/app/components/tabs/CurationToolsTab.tsx
@@ -1,105 +1,162 @@
-import { PAGE_ROUTE } from 'app/config/constants/constants';
-import { Gene, Meta } from 'app/shared/model/firebase/firebase.model';
-import GeneSelect from 'app/shared/select/GeneSelect';
-import { getFirebasePath } from 'app/shared/util/firebase/firebase-utils';
 import { componentInject } from 'app/shared/util/typed-inject';
 import { IRootStore } from 'app/stores';
 import React, { useEffect, useRef, useState } from 'react';
-import { FaExclamationCircle } from 'react-icons/fa';
-import { Button, Col, Row } from 'reactstrap';
-import './curation-tools-tab.scss';
+import { Button, Row } from 'reactstrap';
+import { FaRegCheckCircle } from 'react-icons/fa';
+import { FaRegCircleXmark } from 'react-icons/fa6';
 import { notifyError } from 'app/oncokb-commons/components/util/NotificationUtils';
-import { isPromiseOk } from 'app/shared/util/utils';
-import { getErrorMessage } from 'app/oncokb-commons/components/alert/ErrorAlertUtils';
-import { observer } from 'mobx-react';
+import { IGene } from 'app/shared/model/gene.model';
+import _ from 'lodash';
+import { IFlag } from 'app/shared/model/flag.model';
+import { CURRENT_REVIEWER } from 'app/config/constants/constants';
 
-function CurationToolsTab({ addMetaListener, metaData, createMetaGene, deleteMetaGene, createGene, deleteGene }: StoreProps) {
-  const selectedGene = useRef<string>(null);
-  const [createButtonDisabled, setCreateButtonDisabled] = useState(true);
-  const [showGeneExistsWarning, setShowGeneExistsWarning] = useState(false);
+export type ReleaseGeneTestData = {
+  passed: boolean;
+  text: string;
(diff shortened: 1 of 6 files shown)


src/main/webapp/app/components/tabs/CurationToolsTab.tsx
```diff
@@ -1,105 +1,142 @@
...
+  }
+
+  useEffect(() => {
+    const geneData = geneEntities.find(entity => entity.hugoSymbol === gene.name);
+    setIsReleased(geneData?.flags?.some(flag => isReleasedFlag(flag)) || false);
+    geneToUpdate.current = geneData;
+  }, []);
+
+  function getStatusIcon(checkPassed: boolean) {
+    return checkPassed ? <FaRegCheckCircle className="text-success mt-1 mr-2" /> : <FaRegCircleXmark className="text-danger mt-1 mr-2" />;
```
zhx828, line 48: Let's move the common class names as a constant within the method.

src/main/webapp/app/components/tabs/CurationToolsTab.tsx
```diff
@@ -1,105 +1,142 @@
...
+    return flag.type === 'GENE_PANEL' && flag.flag === 'ONCOKB';
+  }
+
+  useEffect(() => {
+    const geneData = geneEntities.find(entity => entity.hugoSymbol === gene.name);
+    setIsReleased(geneData?.flags?.some(flag => isReleasedFlag(flag)) || false);
+    geneToUpdate.current = geneData;
+  }, []);
+
+  function getStatusIcon(checkPassed: boolean) {
```
zhx828, line 47: Given all use cases within this file, let's rename the pass to checkNotPassed and update the usage.

src/main/webapp/app/components/tabs/CurationToolsTab.tsx
```diff
@@ -1,105 +1,142 @@
...
+            Gene summary is not empty
+          </Row>
+          <Row className="mb-1">
+            {getStatusIcon(!geneBackgroundEmpty)}
+            Gene background is not empty
+          </Row>
+          <Row className="mb-3">
+            {getStatusIcon(!geneTypeEmpty)}
+            Gene type is specified
+          </Row>
```
zhx828, line 105: Can we aggregate these checks into a list? The `Row` here can be populated based on the list. Also `confirmButtonDisabled` can be calculated based on the list. 
Test needs to be differentiated with `optional` and `requried`

zhx828 (changes requested):
- Could you also update the title to reflect the purpose of the PR? Also make sure capitalize the First word so it can be used in the release note. We don't generate release note yet in this repo, but good to have it standardized.
- Could you add label, `feature`?

src/main/webapp/app/components/tabs/CurationToolsTab.tsx
bprize15: After defining the tests to be run when creating a gene, we now have a `passed` variable for each test. The use cases all now use `getStatusIcon(test.passed)`, so will leave the same

src/main/webapp/app/components/tabs/CurationToolsTab.tsx
```diff
@@ -1,105 +1,159 @@
...
+    },
+    {
+      passed: !!gene.background,
+      text: 'Gene background is not empty',
+      type: 'required',
+    },
+    {
+      passed: !!gene.type.ocg || !!gene.type.tsg,
+      text: 'Gene type is specified',
+      type: 'required',
```
zhx828, line 40: Forgot to mention, the gene type check is optional.

src/main/webapp/app/components/tabs/CurationToolsTab.tsx
```diff
@@ -1,105 +1,159 @@
...
+  text: string;
+  type: 'optional' | 'required';
+};
+
+export function CurationToolsTab({ metaList, addMetaListListener, gene, geneEntities, searchGenes, updateGene, searchFlags }: StoreProps) {
+  const reviewList = metaList?.[gene.name].review;
+
+  const tests: ReleaseGeneTestData[] = [
+    {
+      passed: reviewList ? !Object.keys(reviewList).some(key => key !== 'currentReviewer' && reviewList[key]) : true,
```
zhx828, line 23: Can we create a constant of `currentReviewer`? @calvinlu3 this is already related to ur work I believe on how to determine whether there is review content.

zhx828 (commented):
Two last comments, once fixed, you can merge. Thanks!

PR #172, opened 2024-01-22
Add drug synonym to dropdown

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/232


zhx828:
@bprize15  Two things before looking to the code
- bug: the drugs api call is made x number of times based on x number of Therapy select is available in the modal
- I do expect to see more synonyms if the space permits
Screenshot 2024-01-22 at 12 26 21 PM

bprize15:
> @bprize15 Two things before looking to the code
> 
> * bug: the drugs api call is made x number of times based on x number of Therapy select is available in the modal
> * I do expect to see more synonyms if the space permits
> 
> Screenshot 2024-01-22 at 12 26 21 PM

Last thing regarding your comment about each dropdown making an API call. I have addressed this by allowing the drugs to optionally be passed in. If they are, the dropdown doesn't make a call.
Currently the therapy modal is receiving its drugList from the curation page and passing it further to its dropdowns. Would you like to keep it that way or have the drugs fetched when the modal is opened?

PR #174, opened 2024-01-25
Update alteration parser logics

zhx828:
There are more improvements we can add. But this is the one to fix a 500 error.

### DIFF
--- a/src/main/java/org/mskcc/oncokb/curation/service/MainService.java
+++ b/src/main/java/org/mskcc/oncokb/curation/service/MainService.java
@@ -88,31 +88,7 @@ public class MainService {
         return Optional.empty();
     }
 
-    private void enrichQuery(Alteration alteration) {
-        if (alteration.getGenes().size() > 0) {
-            alteration.setGenes(
-                alteration
-                    .getGenes()
-                    .stream()
-                    .map(gene -> {
-                        if (gene.getId() != null) {
-                            Optional<Gene> geneOptional = geneService.findOne(gene.getId());
-                            if (geneOptional.isEmpty()) {
-                                return gene;
-                            } else {
-                                return geneOptional.get();
-                            }
-                        }
-                        return gene;
-                    })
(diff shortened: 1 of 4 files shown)


src/main/java/org/mskcc/oncokb/curation/service/MainService.java
```diff
@@ -110,6 +140,7 @@ public void annotateAlteration(Alteration alteration) {
             alteration.setGenes(pcAlteration.getGenes());
         }
         alteration.setAlteration(pcAlteration.getAlteration());
+        alteration.setName(pcAlteration.getName());
```
bprize15, line 143: Line 197 has duplicate logic with a check. Would remove one or the other depending how confident you are alteration name won't be null.

src/main/java/org/mskcc/oncokb/curation/service/MainService.java
```diff
@@ -85,7 +88,31 @@ public Optional<Sequence> findSequenceByGene(ReferenceGenome referenceGenome, In
...
+                        }
+                        return gene;
+                    })
+                    .collect(Collectors.toSet())
+            );
+        }
+    }
+
     public void annotateAlteration(Alteration alteration) {
+        enrichQuery(alteration);
```
bprize15, line 114: Logic on line 147 seems to do same thing. Do we need both?

src/main/java/org/mskcc/oncokb/curation/service/MainService.java
zhx828: That's to use protein change as name.

src/main/java/org/mskcc/oncokb/curation/service/MainService.java
zhx828: Good call. Updated the logics.

bprize15: Sounds good otherwise LGTM

PR #176, opened 2024-01-31
Add additional germline features

zhx828:
- [Update curation germline features](https://github.com/oncokb/oncokb-transcript/commit/809c1a532d62146ca7cfc8c5b24094bccfb03b9f)
- [Update entities for GenomicIndicator integration && Add AlleleState](https://github.com/oncokb/oncokb-transcript/commit/ccbbd87413c0acf3212ee0d8774f631d1ee43b0a)
- Move gene info next to the gene name in the curation page

This fixes https://github.com/oncokb/oncokb-pipeline/issues/237 and https://github.com/oncokb/oncokb-pipeline/issues/242

### DIFF
--- a/src/main/webapp/app/entities/genomic-indicator/genomic-indicator.tsx
+++ b/src/main/webapp/app/entities/genomic-indicator/genomic-indicator.tsx
@@ -18,11 +18,13 @@ import { IAlleleState } from 'app/shared/model/allele-state.model';
 export interface IGenomicIndicatorProps extends StoreProps, RouteComponentProps<{ url: string }> {}
 
 const getAssociatedAlterations = (associations?: IAssociation[]) => {
-  return associations
-    ?.map(association =>
-      association.alterations?.map(alt => `${alt.genes?.map(gene => gene.hugoSymbol).join('-')} ${alt.alteration}`).join()
-    )
-    .join();
+  return (
+    associations
+      ?.map(association =>
+        association.alterations?.map(alt => `${alt.genes?.map(gene => gene.hugoSymbol).join('-')} ${alt.alteration}`).join()
+      )
+      .join() || ''
+  );
 };
 
 const getAssociatedAlleleStates = (alleleStates?: IAlleleState[]) => {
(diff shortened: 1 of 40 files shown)


src/main/webapp/app/entities/genomic-indicator/genomic-indicator.tsx
```diff
@@ -7,19 +7,60 @@ import { ENTITY_ACTION, ENTITY_TYPE } from 'app/config/constants/constants';
...
+    {
+      Header: 'Associated Allele States',
+      Cell(cell: { original }): JSX.Element {
+        return <span>{getAssociatedAlleleStates(cell.original.alleleStates)}</span>;
+      },
+      onFilter: (data: IGenomicIndicator, keyword) =>
+        data.alleleStates ? filterByKeyword(getAssociatedAlleleStates(data.alleleStates), keyword) : false,
+    },
+    {
+      Header: 'Associated Alterations',
```
calvinlu3, line 57: Should add `id` when we don't have `accessor`.

src/main/webapp/app/shared/model/firebase/firebase.model.ts
```diff
@@ -109,6 +109,14 @@ export class Treatment {
   }
 }
 
+type AlleleState = 'monoallelic' | 'biallelic' | 'mosaic';
+
+export class GenomicIndicator {
+  indicator = '';
```
calvinlu3, line 115: Do we want to review these fields?

zhx828: yeah, for firebase model

src/main/webapp/app/entities/genomic-indicator/genomic-indicator.tsx
zhx828: oh, ok. Will do. Is it required?

calvinlu3: Only required if accessor is not defined. I believe the sorting will also not work for `Associated Allele States` and `Associated Alterations` because the sorting by default will try to use the value that is found using the accessor.

zhx828: Added the accessor so it can be sorted.

calvinlu3 (approved):
LGTM

PR #177, opened 2024-01-31
Add review page

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3646


zhx828:
The feature overall looks good. Two minor fixes needed before code review
- after deleting Tx. The change in the review is not expected PTEN/Oncogenic Mutations/Capivasertib + Fulvestrant
Screenshot 2024-02-01 at 11 48 23 AM
- only bold the title when it's not `disableLeftBorder`

calvinlu3:
The Tx level should be fixed now
image

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -239,271 +250,333 @@ const CurationPage = (props: ICurationPageProps) => {
...
+      }
+    } else {
+      if (isReviewFinished) {
+        button = (
+          <Button color="primary" onClick={handleReviewButtonClick}>
+            Review Complete
+          </Button>
+        );
+      } else {
+        return undefined;
```
zhx828, line 288: Can this really happen?

zhx828 (approved):
I scanned the PR and didn't see anywhere needs to change right away. We will add a lot of tests for this component, we will revisit again. Thank you for your hard work on this component!

src/main/webapp/app/pages/curation/CurationPage.tsx
calvinlu3: `isReviewFinished` is false by default (and only gets set in the review mode) and when a gene doesn't need review, this condition will be hit. We don't want to show a button in this scenario.

PR #178, opened 2024-01-31
Enable comments on Curation Page Collapsible/Sections

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/238

### DIFF
--- a/src/main/webapp/app/pages/curation/CurationPage.tsx
+++ b/src/main/webapp/app/pages/curation/CurationPage.tsx
@@ -17,7 +17,7 @@ import {
 } from 'app/config/constants/constants';
 import { PubmedGeneLink } from 'app/shared/links/PubmedGeneLink';
 import { InlineDivider, PubmedGeneArticlesLink } from 'app/shared/links/PubmedGeneArticlesLink';
-import { getSectionClassName } from 'app/shared/util/utils';
+import { getSectionClassName, getUserFullName } from 'app/shared/util/utils';
 import ExternalLinkIcon from 'app/shared/icons/ExternalLinkIcon';
 import WithSeparator from 'react-with-separator';
 import { AutoParseRefField } from 'app/shared/form/AutoParseRefField';
@@ -25,7 +25,8 @@ import { RealtimeCheckedInputGroup, RealtimeTextAreaInput } from 'app/shared/fir
 import { getFirebasePath, getMutationName } from 'app/shared/util/firebase/firebase-utils';
 import styles from './styles.module.scss';
 import { notifyError } from 'app/oncokb-commons/components/util/NotificationUtils';
-import { HistoryRecord, Mutation } from 'app/shared/model/firebase/firebase.model';
+import { CancerType, Comment, DX_LEVELS, HistoryRecord, Mutation, PX_LEVELS, TX_LEVELS } from 'app/shared/model/firebase/firebase.model';
+import RealtimeDropdownInput from 'app/shared/firebase/input/RealtimeDropdownInput';
 import { GENE_TYPE, GENE_TYPE_KEY, MUTATION_EFFECT_OPTIONS, ONCOGENICITY_OPTIONS, TX_LEVEL_OPTIONS } from 'app/config/constants/firebase';
 import GeneHistoryTooltip from 'app/components/geneHistoryTooltip/GeneHistoryTooltip';
 import VusTable from '../../shared/table/VusTable';
@@ -38,6 +39,7 @@ import MutationCollapsible from './collapsible/MutationCollapsible';
 import { IDrug } from 'app/shared/model/drug.model';
 import { IGene } from 'app/shared/model/gene.model';
 import CurationToolsTab from 'app/components/tabs/CurationToolsTab';
+import CommentIcon from 'app/shared/icons/CommentIcon';
 import { HgncLink } from 'app/shared/links/HgncLink';
 
 export interface ICurationPageProps extends StoreProps, RouteComponentProps<{ hugoSymbol: string }> {}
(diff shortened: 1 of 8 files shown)


src/main/webapp/app/shared/icons/CommentIcon.tsx
```diff
@@ -156,35 +161,44 @@ const CommentBox = observer((props: ICommentBoxProps) => {
   );
 });
 
+/* eslint-disable no-console */
```
zhx828, line 164: Needs to be removed.

src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
```diff
@@ -46,11 +47,53 @@ const MutationCollapsible = ({
...
   drugList,
+  handleFirebaseUpdateUntemplated,
+  handleFirebaseDeleteFromArray,
+  account,
 }: IMutationCollapsibleProps) => {
   const title = getMutationName(mutation);
   const mutationFirebasePath = buildFirebaseGenePath(hugoSymbol, `mutations/${firebaseIndex}`);
   const showMutationLevelSummary = !title.includes(',');
 
+  async function handleCreateComment(path: string, content: string, currentCommentsLength: number) {
```
zhx828, line 58: Is this the same as the one in CurationPage?

src/main/webapp/app/shared/model/firebase/firebase.model.ts
```diff
@@ -80,7 +80,7 @@ export const TI_NAME: { [key in TI_TYPE]: string } = {
 };
 
 // We should convert boolean strings to actual booleans
-export type BoolString = 'false' | 'true';
+export type BoolString = 'false' | 'true' | boolean;
```
zhx828, line 83: The type is for a string boolean though.

src/main/webapp/app/shared/util/firebase/firebase-crud-store.ts
```diff
@@ -70,14 +71,23 @@ export class FirebaseCrudStore<T> {
     }
   }
 
+  async updateUntemplated(path: string, value: any) {
```
zhx828, line 74: What does this name mean?

src/main/webapp/app/shared/util/firebase/firebase-crud-store.ts
```diff
@@ -70,14 +71,23 @@ export class FirebaseCrudStore<T> {
...
+
   delete(path: string) {
     return remove(ref(this.db, path));
   }
 
   async deleteFromArray(path: string, indices: number[]) {
     return await runTransaction(ref(this.db, path), (currentData: any[]) => {
       const newData = [];
-      for (let i = 0; i < currentData.length; i++) {
+      for (let i = 0; currentData !== null && i < currentData.length; i++) {
```
zhx828, line 90: In what scenario this will ben null?

zhx828 (changes requested):

src/main/webapp/app/shared/util/firebase/firebase-crud-store.ts
bprize15: The update function currently used is templated depending on the store. For example, if we have `FireBaseCrudStore<Gene>`, then the update function looks like `update(path: string, value: Gene)`. I think it is useful to have an untemplated version so we can write anywhere without requiring a store.

src/main/webapp/app/shared/model/firebase/firebase.model.ts
bprize15: Will change `name_comments`' `resolved` field so that it is of type `BoolString | boolean` instead

src/main/webapp/app/shared/util/firebase/firebase-crud-store.ts
bprize15: Honestly, I do not remember. I think I must've encountered it for me to have put that there. However, I can think of a scenario off the top of my head:

Since it uses `runTransaction` any changes to the currentdata will result in the transaction being rerun with updated currentdata. Thus, if someone deletes all comments while you are deleting a comment, currentData can become null when the function is rerun. 

Wanted to get this comment in before the end of the day since you won't be in tomorrow, but I will try to reproduce to remind myself why I originally put that there.

src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
bprize15: Yeah it's the same but not sure if we wanted to couple the behavior. Figured it might be safer to have 2 copies, in case either needed additional logic, but I can pass down from the Curation Page if you prefer

zhx828 (changes requested):

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -244,63 +285,113 @@ const CurationPage = (props: ICurationPageProps) => {
...
+                <CommentIcon
+                  id={`${hugoSymbol}_curation_page`}
+                  comments={props.data.name_comments || []}
+                  onCreateComment={content =>
+                    handleCreateComment(`${firebaseGenePath}/name_comments`, content, props.data.name_comments?.length || 0)
+                  }
+                  onDeleteComments={indices => handleDeleteComments(`${firebaseGenePath}/name_comments`, indices)}
+                  onResolveComment={index => handleResolveComment(`${firebaseGenePath}/name_comments/${index}`)}
+                  onUnresolveComment={index => handleUnresolveComment(`${firebaseGenePath}/name_comments/${index}`)}
+                />
```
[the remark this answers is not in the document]
bprize15: Yeah that makes sense, will make those changes

src/main/webapp/app/shared/model/firebase/firebase.model.ts
```diff
@@ -165,7 +168,10 @@ export class Mutation {
   germline_genomic_indicators: GenomicIndicator[] = [];
   mutation_effect: MutationEffect = new MutationEffect();
   mutation_effect_uuid: string = generateUuid();
+  mutation_effect_comments?: Comment[] = []; // used for somatic
+  mutation_effect_germline_comments?: Comment[] = [];
```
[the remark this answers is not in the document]
bprize15: I checked an the old Curation Platform pulls from this location (since I can see old comments in the new platform which reads from this location). I checked BRAF, V600 to verify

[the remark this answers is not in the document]
zhx828: @bprize15 the old curation platform doesn't have germline integrated.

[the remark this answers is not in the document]
bprize15: `mutation_effect_comments` was in this location, so I was able to put `mutation_effect_germline_comments` wherever I wanted. I figured it made the most sense to keep it in the same object as `mutation_effect_comments`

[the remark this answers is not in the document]
zhx828: I figured it would makes sense to leave the comments on the same level of germline which is under mutaiton_effect. We can merge as it and come back to update when needed.

zhx828 (approved):

PR #190, opened 2024-02-08
Make mutations scrollable

bprize15:
Change addresses https://github.com/oncokb/oncokb-pipeline/issues/248

### DIFF
--- a/src/main/webapp/app/shared/table/VusTable.tsx
+++ b/src/main/webapp/app/shared/table/VusTable.tsx
@@ -232,7 +232,7 @@ const VusTable = ({
   return vusData ? (
     <div className={'justify-content-between align-items-center mt-4'}>
       <div className={'d-flex align-items-baseline'}>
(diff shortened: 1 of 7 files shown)


src/main/webapp/app/shared/table/VusTable.tsx
```diff
@@ -232,7 +232,7 @@ const VusTable = ({
   return vusData ? (
     <div className={'justify-content-between align-items-center mt-4'}>
       <div className={'d-flex align-items-baseline'}>
-        <span style={{ fontSize: '2rem' }}>Variants of Unknown Significance (Investigated and data not found)</span>
+        <h5>Variants of Unknown Significance (Investigated and data not found)</h5>
```
zhx828, line 235: I want to keep the button on the same level of the header. Let's adjust fontsize instead

zhx828 (approved):

PR #195, opened 2024-02-13
Improve alteration modal

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/252
resolves https://github.com/oncokb/oncokb-pipeline/issues/253

### DIFF
--- a/src/main/webapp/app/shared/modal/AddMutationModal.tsx
+++ b/src/main/webapp/app/shared/modal/AddMutationModal.tsx
@@ -1,49 +1,41 @@
 import Tabs from 'app/components/tabs/tabs';
 import React, { KeyboardEventHandler, useCallback, useEffect, useMemo, useRef, useState } from 'react';
-import { Alert, Button, Col, Modal, ModalBody, ModalFooter, Row } from 'reactstrap';
+import { Alert, Button, Col, Input, Modal, ModalBody, ModalFooter, Row } from 'reactstrap';
 import CreatableSelect from 'react-select/creatable';
-import ValidatedForm from 'app/shared/form/ValidatedForm';
-import { ValidatedField, ValidatedSelect } from 'app/shared/form/ValidatedField';
-import { notifyError } from 'app/oncokb-commons/components/util/NotificationUtils';
 import _ from 'lodash';
-import { generateUuid, parseAlterationName } from '../util/utils';
+import { notNullOrUndefined, parseAlterationName } from '../util/utils';
 import { IRootStore } from 'app/stores';
 import { componentInject } from '../util/typed-inject';
 import { flow, flowResult } from 'mobx';
-import { CancellablePromise } from 'mobx/dist/internal';
-import { EntityStatusAlteration } from '../api/generated';
+import { AlterationTypeEnum, EntityStatusAlteration, Gene } from '../api/generated';
 import './add-mutation-modal.scss';
 import { IGene } from '../model/gene.model';
-import { FaChevronDown, FaChevronUp, FaExclamationTriangle } from 'react-icons/fa';
+import { FaChevronDown, FaChevronUp, FaExclamationCircle, FaExclamationTriangle } from 'react-icons/fa';
 import { Alteration, Mutation } from '../model/firebase/firebase.model';
+import ReactSelect, { MenuPlacement } from 'react-select';
+import { notifyError } from 'app/oncokb-commons/components/util/NotificationUtils';
+import styles from './styles.module.scss';
 
-// Follow naming convention: all words start with uppercase, and space are denoted by _
-// This is used to populate field names
-type MutationState = {
-  label: string;
-  value: string;
-  uuid: string;
-  Alteration: string;
-  Name: string;
-  Protein_Change: string;
-  Protein_Start: string;
-  Protein_End: string;
-  Ref_Residues: string;
-  Var_Residues: string;
-  Consequence: string;
-  Comment?: string;
-  Excluding: MutationState[];
+type AlterationData = {
+  type: AlterationTypeEnum;
+  alteration: string;
+  name: string;
+  consequence: string;
+  comment: string;
+  excluding: AlterationData[];
+  genes?: Gene[];
+  proteinChange?: string;
+  proteinStart?: number;
+  proteinEnd?: number;
+  refResidues?: string;
+  varResidues?: string;
   warning?: string;
   error?: string;
-};
-
-type MutationName = {
-  name: string;
-  excluding: string[];
-  comment: string;
+  alterationFieldValueWhileFetching?: string;
 };
 
 interface IAddMutationModalProps extends StoreProps {
+  mutationList: Mutation[];
   hugoSymbol: string;
   isOpen: boolean;
   onConfirm: (alterations: Alteration[]) => void;
@@ -58,6 +50,7 @@ function AddMutationModal(props: IAddMutationModalProps) {
 }
 
 function AddMutationModalContent({
+  mutationList,
   hugoSymbol,
   mutationToEdit,
   annotateAlteration,
@@ -67,11 +60,20 @@ function AddMutationModalContent({
   onConfirm,
   onCancel,
 }: Omit<IAddMutationModalProps, 'isOpen'>) {
-  const consequenceOptions = consequences.map(consequence => ({ label: consequence.name, value: consequence.id }));
+  const typeOptions: DropdownOption[] = [
+    AlterationTypeEnum.ProteinChange,
+    AlterationTypeEnum.CopyNumberAlteration,
+    AlterationTypeEnum.StructuralVariant,
+    AlterationTypeEnum.CdnaChange,
+    AlterationTypeEnum.Any,
+  ].map(type => ({ label: type, value: type }));
+  const consequenceOptions: DropdownOption[] = consequences.map(consequence => ({ label: consequence.name, value: consequence.id }));
 
   const [inputValue, setInputValue] = useState('');
-  const [mutationStates, setMutationStates] = useState<MutationState[]>([]);
+  const [tabStates, setTabStates] = useState<AlterationData[]>([]);
+  const [excludingInputValue, setExcludingInputValue] = useState('');
   const [excludingCollapsed, setExcludingCollapsed] = useState(true);
+  const [mutationAlreadyExists, setMutationAlreadyExists] = useState(false);
 
   const inputRef = useRef(null);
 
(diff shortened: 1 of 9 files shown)


zhx828 (changes requested):
- When adding an alteration that already in the dropdown, it should dismiss the second one.
- Type should be shown above alteration
- Consequence should only be visible when it's protein change
- Currently the excluding components only show up if the curator knows how to write the excluding in the alteration. Is there a way to allow adding excluding within the tab content?

src/main/webapp/app/shared/util/utils.tsx
```diff
@@ -205,3 +269,7 @@ export function findIndexOfFirstCapital(str: string) {
 export function isNumeric(value: string) {
   return /^-?\d+$/.test(value);
 }
+
+export function notNullOrUndefined(val) {
```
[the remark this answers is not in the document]
bprize15: For all my use cases, it has been a number, but I figured it made sense to generalize since this could be used for any type.

src/main/webapp/app/shared/util/utils.tsx
```diff
@@ -176,23 +176,87 @@ export async function isPromiseOk(promise: Promise<any>) {
...
+  const alleles = rest ? rest.split('/') : [];
+  for (const allele of alleles) {
+    alterations.push(firstPart + allele.trim());
   }
   return alterations;
 }
 
+// splits alteration name into alteration, excluding and comment
+// if alteration is separated by /, applies the same excluding and comment to separated alterations
+export function parseAlterationName(alterationName: string): { name: string; excluding: string[]; comment: string; variantName: string }[] {
```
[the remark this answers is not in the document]
bprize15: What we are referring to as name here is actually the alteration, and variantName is supposed to be referring to name (matching with the modal). I have updated both so that they match with the modal.

zhx828 (commented):
- when modifying the mutation, the `Mutation already exists` always show up
- The button text should be `Update` instead of `Add` when modifying the mutation
- if alteration is in the mutation string with other alterations and it has a dedicated entry within the mutation list, it should not be error `Mutation already exists` but rather a warning `Mutation already exists`
- when there is no alteration specified in the Excluding clause, the arrow-down icon is confusing. I kept clicking on it without know it's disabled. Maybe don't show the icon until it's not empty?
- We should support `Any` in the alteration type as well.

bprize15:
> * when modifying the mutation, the `Mutation already exists` always show up
> * The button text should be `Update` instead of `Add` when modifying the mutation
> * if alteration is in the mutation string with other alterations and it has a dedicated entry within the mutation list, it should not be error `Mutation already exists` but rather a warning `Mutation already exists`
> * when there is no alteration specified in the Excluding clause, the arrow-down icon is confusing. I kept clicking on it without know it's disabled. Maybe don't show the icon until it's not empty?
> * We should support `Any` in the alteration type as well.

Based on our discussion, I have kept the `Mutation already exists` as an error instead of a warning for the time being. All other changes were implemented

src/main/webapp/app/shared/util/utils.tsx
```diff
@@ -176,23 +176,87 @@ export async function isPromiseOk(promise: Promise<any>) {
...
 
+// splits alteration name into alteration, excluding and comment
+// if alteration is separated by /, applies the same excluding and comment to separated alterations
+export function parseAlterationName(alterationName: string): { alteration: string; excluding: string[]; comment: string; name: string }[] {
+  let regex = new RegExp('\\[(.*)\\]', 'i');
+  const nameSection = regex.exec(alterationName);
+  let name = '';
+  if (nameSection?.length > 1) {
+    name = nameSection[1];
+  }
```
zhx828, line 207: What's the logic here?

src/main/webapp/app/shared/modal/AddMutationModal.tsx
```diff
@@ -80,507 +82,747 @@ function AddMutationModalContent({
...
-    function replaceOldMutations(oldMutationStates: MutationState[], newMutations: MutationState[], uuidToReplace: string) {
-      return oldMutationStates.reduce((accumulator: MutationState[], currentMutation) => {
-        if (uuidToReplace === currentMutation.uuid) {
-          return [...accumulator, ...newMutations];
-        }
-        return [...accumulator, currentMutation];
-      }, []);
+  async function fetchExcludedAlteration(
+    newAlteration: string,
+    tabIndex: number,
```
zhx828, line 329: I'm a bit confused through out this file because of this naming. The tab concept is also used in excluding. Should we rename to alterationIndex?

src/main/webapp/app/shared/modal/AddMutationModal.tsx
```diff
@@ -80,507 +82,747 @@ function AddMutationModalContent({
...
-          Excluding: [],
-          warning: excludingAlteration.warning ? excludingAlteration.message : null,
-          error: excludingAlteration.error ? excludingAlteration.message : null,
-        })),
-        warning: alteration.warning ? alteration.message : null,
-        error: alteration.error ? alteration.message : null,
-      };
+  function handleNormalAlterationChange(newValue: string, tabIndex: number) {
+    setTabStates(states => {
+      const newStates = _.cloneDeep(states);
```
zhx828, line 320: Is it necessary to do deep clone?

src/main/webapp/app/shared/modal/AddMutationModal.tsx
```diff
@@ -80,507 +82,747 @@ function AddMutationModalContent({
...
 
-    setMutationStates(prev => {
-      if (excludingUuid) {
-        const newMutationStates = _.cloneDeep(prev);
-        const parentState = newMutationStates.find(state => state.uuid === uuid);
-        newAlterations.forEach(newAlteration => (newAlteration.Comment = null));
-        parentState.Excluding = replaceOldMutations(parentState.Excluding, newAlterations, excludingUuid);
-        return newMutationStates;
-      }
+    const newComment = newParsedAlteration[0].comment;
```
zhx828, line 362: Is this needed? At line 378, we could just use `newParsedAlteration[index].comment`?

src/main/webapp/app/shared/modal/AddMutationModal.tsx
```diff
@@ -80,507 +82,747 @@ function AddMutationModalContent({
...
-      if (uuid) {
-        return replaceOldMutations(prev, newAlterations, uuid);
-      }
+    const alterationPromises: Promise<EntityStatusAlteration>[] = [];
+    let newAlterations: AlterationData[] = [];
+    if (newParsedAlteration[0].alteration !== alterationData[tabIndex]?.excluding[excludingIndex].alteration) {
+      alterationPromises.push(fetchAlteration(newParsedAlteration[0].alteration));
+    } else {
+      newAlterations.push(alterationData[tabIndex].excluding[excludingIndex]);
+    }
```
zhx828, line 370: I'm a bit lost here. Could you explain what's the logic here?

src/main/webapp/app/shared/modal/AddMutationModal.tsx
```diff
@@ -80,507 +82,747 @@ function AddMutationModalContent({
...
     if (!inputValue) return;
     if (event.key === 'Enter' || event.key === 'tab') {
-      await fetchAlteration(inputValue);
+      const newParsedAlteration = filterAlterationsAndNotify(parseAlterationName(inputValue), tabStates);
+      if (newParsedAlteration.length === 0) {
+        return;
+      }
+
+      const newComment = newParsedAlteration[0].comment;
+      const newVariantName = newParsedAlteration[0].name;
```
zhx828, line 444: Same here. Doesn't seem to me it's necessary to create two vars here.

src/main/webapp/app/shared/modal/AddMutationModal.tsx
```diff
@@ -80,507 +82,747 @@ function AddMutationModalContent({
...
+
+      const newEntityStatusAlterationsPromise = fetchAlterations(newParsedAlteration.map(alt => alt.alteration));
+      const newEntityStatusExcludingAlterationsPromise = fetchAlterations(newParsedAlteration[0].excluding);
```
zhx828, line 447: Why we only fetch newParsedAlteration[0]?

src/main/webapp/app/shared/modal/AddMutationModal.tsx
```diff
@@ -80,507 +82,747 @@ function AddMutationModalContent({
...
+      setTabStates(states => [...states, ...newAlterations]);
 
       setInputValue('');
       event.preventDefault();
     }
   };
 
-  function getTabTitle(mutationState: MutationState) {
-    if (!mutationState) {
+  const handleKeyDownExcluding = async (event: React.KeyboardEvent<Element>, tabIndex: number) => {
```
zhx828, line 473: Isn't this method almost identical with `handleKeyDown`?

zhx828 (changes requested):
Sorry about the delay. AddMutationModal.tsx is a big file that I might have to go back and forth with you few times.

src/main/webapp/app/shared/util/utils.tsx
bprize15: This is to check for the variant name in brackets. I use a regex, and if there is a match, the length should be two. Index 0 will be the whole match, and index 1 will be the captured group (the variant name)

src/main/webapp/app/shared/modal/AddMutationModal.tsx
bprize15: Yeah will definitely do that. I see how it's confusing,

src/main/webapp/app/shared/modal/AddMutationModal.tsx
bprize15: `newParsedAlteration` only is an array because the user can enter something like `V600E/G` which would result in two alterations. However, I assume that if you write `V600E/G {excluding V600}`, the excluding applies to both, so we only need to fetch the excluding alterations once.

src/main/webapp/app/shared/modal/AddMutationModal.tsx
bprize15: I do the check to save a fetch for the following case. You start with the alteration `V600E`. You then change the alteration to `V600E/G`. Since the resulting alterations are `V600E` and `V600G` there is no need to refetch the first alteration since we already have it. In that case, I just added it to new alterations. 

If the alteration changed from `V600E/G` to `V600F/G` the logic in the if branch will execute since the first alteration would need to be refetched.

src/main/webapp/app/shared/modal/AddMutationModal.tsx
bprize15: You are correct. Either way would be correct since the only way multiple alterations would exist is if the alteration changed from `V600E` to `V600E/G (comment)`. Since the comment would apply to both, I chose to use the first comment, but both would be the same anyway.

bprize15: Changing it to just use index

src/main/webapp/app/shared/modal/AddMutationModal.tsx
bprize15: Correct, for same reasoning as above. Will change to use index.

src/main/webapp/app/shared/modal/AddMutationModal.tsx
bprize15: I think there are enough differences to justify two different methods. The `setState` calls are different, `handleKeyDown` needs to fetch excluding and non-excluding while `handleKeyDownExcluding` only needs to fetch its own alterations, and `handleKeyDownExcluding` handles checking for duplicate alterations differently.

I think there are a lot of cases where the logic for normal alterations and excluding alterations appear similar, but they are just different enough to make abstracting the logic messy. I tried to abstract a bunch the first time I created this component, and I think the resulting code was harder to maintain.

src/main/webapp/app/shared/modal/AddMutationModal.tsx
bprize15: React recommends treating state as immutable and always making copies: https://react.dev/learn/updating-objects-in-state.

I also just tested, and my component was not rerendering when I mutated the state. I needed to return a new object with a new reference.

src/main/webapp/app/shared/util/utils.tsx
zhx828: oh right, I forgot about the first element is the whole match. Thanks.

zhx828 (approved):
LGTM, thanks.

PR #198, opened 2024-02-19
Add rct modal

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/229

### DIFF
--- a/src/main/java/org/mskcc/oncokb/curation/service/CancerTypeService.java
+++ b/src/main/java/org/mskcc/oncokb/curation/service/CancerTypeService.java
@@ -1,9 +1,24 @@
 package org.mskcc.oncokb.curation.service;
 
+import com.mysql.cj.util.StringUtils;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
 import org.mskcc.oncokb.curation.domain.CancerType;
+import org.mskcc.oncokb.curation.domain.RelevantCancerTypeBody;
+import org.mskcc.oncokb.curation.domain.RelevantCancerTypeQuery;
+import org.mskcc.oncokb.curation.domain.enumeration.TumorForm;
 import org.mskcc.oncokb.curation.repository.CancerTypeRepository;
+import org.mskcc.oncokb.curation.util.CancerTypeUtils;
+import org.mskcc.oncokb.curation.util.enumeration.RelevantCancerTypeDirection;
+import org.mskcc.oncokb.curation.util.enumeration.SpecialCancerType;
+import org.oncokb.client.Evidence.LevelOfEvidenceEnum;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.data.domain.Page;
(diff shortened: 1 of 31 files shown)


zhx828:
- if you change the line height to 1.2 and font size to 0.7rem, it should be similar size to other action icons.
- if we want to make the modal more compact, we can use 5 rows per page and disable the page size dropdownx

bprize15:
> * if you change the line height to 1.2 and font size to 0.7rem, it should be similar size to other action icons.
> * if we want to make the modal more compact, we can use 5 rows per page and disable the page size dropdownx

Made both changes, just wondering about the following:

- The RCT text seems like it might be difficult to read
- For the ones with longer RCTs, will a page size of 5 be inconvenient?

Wondering your thoughts, and you can look at the new version with the changes with the most recent commit

src/main/webapp/app/shared/modal/relevant-cancer-types-modal-store.ts
```diff
@@ -0,0 +1,87 @@
...
+    this.tumor = null;
+    this.level = null;
+  }
+
+  setRelevantCancerTypes(relevantCancerTypes: RelevantCancerType[]) {
+    this.relevantCancerTypes = relevantCancerTypes;
+  }
+
+  setDeleted(index: number, isDeleted: boolean) {
+    /* eslint-disable no-console */
```
zhx828, line 82: let's remove.

src/main/webapp/app/shared/modal/RelevantCancerTypesModal.tsx
```diff
@@ -0,0 +1,241 @@
...
+          }
+          return convertFetchedCancerTypeToRelevantCancerType(rct, true);
+        });
+        relevantCancerTypesModalStore.setRelevantCancerTypes(rcts);
+      }
+
+      if (cancerTypeClient) {
+        setRelevantCancerTypes();
+      }
+    }, [cancerTypeClient]);
```
zhx828, line 80: This is a very interesting way. Would cancerTypeClient ever change?

src/main/java/org/mskcc/oncokb/curation/web/rest/CancerTypeResource.java
```diff
@@ -217,4 +218,12 @@ public ResponseEntity<List<CancerType>> searchCancerTypes(@RequestParam String q
         HttpHeaders headers = PaginationUtil.generatePaginationHttpHeaders(ServletUriComponentsBuilder.fromCurrentRequest(), page);
         return ResponseEntity.ok().headers(headers).body(page.getContent());
     }
+
```
zhx828, line 221: Let's add a blurb describing this method, especially describe all params.

src/main/java/org/mskcc/oncokb/curation/service/CancerTypeService.java
```diff
@@ -128,6 +142,202 @@ public Optional<CancerType> findByMainTypeAndSubtypeIsNull(String mainType) {
         return cancerTypeRepository.findByMainTypeAndSubtypeIsNull(mainType);
     }
 
+    @Transactional(readOnly = true)
+    public List<CancerType> getRelevantCancerTypes(String levelOfEvidence, List<RelevantCancerTypeQuery> cancerTypeQueries) {
+        boolean isLevelBased = levelOfEvidence != null;
+        RelevantCancerTypeDirection direction = levelOfEvidence != null && levelOfEvidence.equals(LevelOfEvidenceEnum.LEVEL_DX1.getValue())
```
zhx828, line 148: Really just nit picking, since you already defined `isLevelBased`, might as well just use here.

src/main/java/org/mskcc/oncokb/curation/service/CancerTypeService.java
```diff
@@ -128,6 +142,202 @@ public Optional<CancerType> findByMainTypeAndSubtypeIsNull(String mainType) {
...
+            ? RelevantCancerTypeDirection.UPWARD
+            : RelevantCancerTypeDirection.DOWNWARD;
+        Set<CancerType> relevantCancerTypes = cancerTypeQueries
+            .stream()
+            .map(query -> {
+                if (StringUtils.isNullOrEmpty(query.getCode())) {
+                    if (isLevelBased) {
+                        List<CancerType> relevant = cancerTypeRepository.findAll();
+                        if (direction.equals(RelevantCancerTypeDirection.UPWARD)) {
```
zhx828, line 157: `direction` is guaranteed to be none null based on the context. But I typically use `RelevantCancerTypeDirection.UPWARD.equal(direction)`

src/main/java/org/mskcc/oncokb/curation/service/CancerTypeService.java
```diff
@@ -128,6 +142,202 @@ public Optional<CancerType> findByMainTypeAndSubtypeIsNull(String mainType) {
...
+                            relevant = findRelevantCancerTypes(query.getMainType(), true, direction);
+                        } else {
+                            relevant =
+                                relevant
+                                    .stream()
+                                    .filter(type ->
+                                        !StringUtils.isNullOrEmpty(type.getMainType()) && type.getMainType().equals(query.getMainType())
+                                    )
+                                    .collect(Collectors.toList());
+                        }
```
zhx828, line 167: Why not just let findRelevantCancerTypes to handle?

src/main/java/org/mskcc/oncokb/curation/service/CancerTypeService.java
```diff
@@ -128,6 +142,202 @@ public Optional<CancerType> findByMainTypeAndSubtypeIsNull(String mainType) {
...
+        boolean isLevelBased = levelOfEvidence != null;
+        RelevantCancerTypeDirection direction = levelOfEvidence != null && levelOfEvidence.equals(LevelOfEvidenceEnum.LEVEL_DX1.getValue())
+            ? RelevantCancerTypeDirection.UPWARD
+            : RelevantCancerTypeDirection.DOWNWARD;
+        Set<CancerType> relevantCancerTypes = cancerTypeQueries
+            .stream()
+            .map(query -> {
+                if (StringUtils.isNullOrEmpty(query.getCode())) {
+                    if (isLevelBased) {
+                        List<CancerType> relevant = cancerTypeRepository.findAll();
```
zhx828, line 156: This is a very costly call. Do we need to load up the whole list?

src/main/java/org/mskcc/oncokb/curation/service/CancerTypeService.java
```diff
@@ -128,6 +142,202 @@ public Optional<CancerType> findByMainTypeAndSubtypeIsNull(String mainType) {
...
+                case ALL_TUMORS:
+                    relevantCancerTypes.addAll(allOncoTreeTypes);
+                    for (CancerType specialType : specialCancerTypes) {
+                        if (
+                            specialType.getCode().equals(SpecialCancerType.ALL_SOLID_TUMORS.name()) ||
+                            specialType.getCode().equals(SpecialCancerType.ALL_LIQUID_TUMORS.name())
+                        ) {
+                            relevantCancerTypes.add(specialType);
+                        }
+                    }
```
zhx828, line 331: Doesn't allOncoTreeTypes already include the special cancer type?

zhx828 (changes requested):

src/main/webapp/app/shared/modal/RelevantCancerTypesModal.tsx
bprize15: No it should not, I think I originally just put it there since it was a dependency I used, but it's a constant export, so I can't see any scenario where it changes.

src/main/java/org/mskcc/oncokb/curation/service/CancerTypeService.java
bprize15: Fixed

src/main/java/org/mskcc/oncokb/curation/service/CancerTypeService.java
bprize15: Nope, I changed it to be less costly

src/main/java/org/mskcc/oncokb/curation/service/CancerTypeService.java
bprize15: I refactored so I don't use allOncoTreeTypes anymore

zhx828 (changes requested):
- [ ] let's disabled RCT when cancer type level is <=0. Add a tooltip asking curator to modify on the Cancer Type instead.
- [x] RCTs should be disabled when level is null
- [ ] Let's update the following message on where the `excluding cancer types` is. Maybe change to excluding cancer types in the `Cancer Type: ***` section above
Screenshot 2024-02-28 at 10 27 31 AM

src/main/resources/config/liquibase/changelog/20220401174649_added_entity_CancerType.xml
```diff
@@ -40,6 +40,13 @@
             </column>
             <!-- jhipster-needle-liquibase-add-column - JHipster will add columns here -->
         </createTable>
+
+        <loadData
+            file="config/liquibase/data/cancer_type.csv"
+            separator=";"
+            tableName="cancer_type"
+            usePreparedStatements="true">
+        </loadData>
```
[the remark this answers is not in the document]
bprize15: Yeah this is for testing. In this case I just exported that database, but if we cannot do that I suppose we can just some fake data?

src/main/resources/config/liquibase/changelog/20231101214350_added_entity_Synonym.xml
```diff
@@ -10,6 +10,11 @@
         Added the entity Synonym.
     -->
     <changeSet id="20231101214350-1" author="jhipster">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <tableExists tableName="synonym"/>
```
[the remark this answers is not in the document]
bprize15: I wasn't able to find that out. I can take some more time and try to get the `h2 console` working, which might provide some insight. I also tried running `clearChecksum` as suggested by https://stackoverflow.com/questions/66377049/table-already-exists-error-when-replacing-yaml-by-xml, but that didn't work either.

src/test/resources/config/application.yml
```diff
@@ -100,6 +102,14 @@ jhipster:
   mail:
     from: test@localhost
     base-url: http://127.0.0.1:8080
+  security:
+    authentication:
+      jwt:
+        # This token must be encoded using Base64 and be at least 256 bits long (you can type `openssl rand -base64 64` on your command line to generate a 512 bits one)
+        base64-secret: NGUwNTlhNDNhOWQ2ZGJlMmEzODczOWJjYWM2MWY4NDAwOTM4YTBmNGFjM2UxNjczZDU2YWZjZjc3MjdkNmU4YjVhZmIwYzI1NDhjMDFkZThiYmE5Mjc5MjM0MGVhODU5MGFhMTAyOTE0M2I3ODA1MGUzZWFhYWUwNWY2ZDgyNjQ=
```
zhx828, line 109: is it necessary to add to application.yml? Can the test use it's own secret?

src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
```diff
@@ -553,6 +558,26 @@ const MutationCollapsible = ({
...
+                                  tooltipProps={
+                                    cancerTypeContainsSpecialCancerType || treatment.level === 'None'
+                                      ? {
+                                          overlay: cancerTypeContainsSpecialCancerType
+                                            ? tooManyRCTsText
+                                            : 'Add a highest level of evidence to view RCTs',
+                                        }
+                                      : null
+                                  }
+                                  disabled={treatment.level === 'None' || cancerTypeContainsSpecialCancerType}
```
zhx828, line 571: The logic of how the tooltip content rendered and when it's disabled are duplicated twice in below. Let's consolidate.

src/main/resources/config/liquibase/changelog/20220401174649_added_entity_CancerType.xml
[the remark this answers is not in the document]
zhx828: Can the loadData only be used when testing? The cancer_type.csv is subject to change in the future. I don't want to update the data here if possible.

src/main/resources/config/liquibase/changelog/20231101214350_added_entity_Synonym.xml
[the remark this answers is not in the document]
zhx828: It would be nice to find out why it happened. But I also know this post recommend always to add preConditions. The sequence of changeset doesn't seem to be honored always. https://stackoverflow.com/questions/61054021/best-practices-on-liquibase-preconditions

src/test/java/org/mskcc/oncokb/curation/service/CancerTypeServiceIT.java
```diff
@@ -0,0 +1,113 @@
...
+
+        List<CancerType> cancerTypes = cancerTypeService.getRelevantCancerTypes("Level_1", relevantCancerTypeBody);
+        String expectedResult = "Mastocytosis, Cutaneous Mastocytosis, Mast Cell Sarcoma"; // excluding also excludes main type if the same. Do we want this?
+        assertEquals(expectedResult, cancerTypesToString(cancerTypes));
+    }
+
+    @Test
+    public void testFindCancerTypesWithDirection() throws Exception {
+        List<CancerType> cancerTypes = cancerTypeService.findRelevantCancerTypes("CML", false, RelevantCancerTypeDirection.DOWNWARD);
+        String expectedResult = "M:All Liquid Tumors, M:All Tumors, Chronic Myelogenous Leukemia, Chronic Myeloid Leukemia, BCR-ABL1+";
```
zhx828, line 64: I do expected to see matched main type to be sorted at the end after the matched cancer type. Is it not the case?

zhx828 (changes requested):

src/test/java/org/mskcc/oncokb/curation/service/CancerTypeServiceIT.java
bprize15: Since we decided to sort by level, main types will end up at the front. However, we can start with the matched cancer type and then sort by level.

src/test/resources/config/application.yml
bprize15: Since this is the application.yml used for all tests, I think it would make sense for the secret to live here. This secret is also already stored in application-prod.yml and application-dev.yml, so I figured the information wasn't sensitive.

However, I could look into disabling authentication for tests if that is what you think is best.

PR #199, opened 2024-02-20
Show mutation stats except oncogenicity for mutations with multiple alterations and update checkbox styles

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/256 and is related to https://github.com/oncokb/oncokb-pipeline/issues/244

Currently, don't show oncogenicity (but show other stats) when mutation has multiple alterations. I omitted the change to show unknown oncogenicity because I need some logic from Ben's PR to check if the alteration is a positional variant. I can follow up with another PR.

Updated style for checkbox/radios:
- Centered the labels
- Centered the checkbox/radio with the text to the right

I adjusted the checkboxes to go under the label instead of inline. The inline saves 1 line of space, but I didn't think it would make that much of a difference. It also makes it consistent with all the other input types. Let me know if I should revert.
image

### DIFF
--- a/src/main/webapp/app/app.scss
+++ b/src/main/webapp/app/app.scss
@@ -394,3 +394,7 @@ http://stackoverflow.com/questions/23436430/bootstrap-3-input-group-100-width */
 .text-decoration-line-through {
   text-decoration: line-through;
 }
+
+.form-check-input {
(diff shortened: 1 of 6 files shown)


zhx828 (approved):
LGTM

PR #202, opened 2024-02-21
Only show one mutation after expanding collapsible

bprize15:


zhx828 (approved):
It looks good! Thanks!

PR #203, opened 2024-02-21
Fix cancer type modal

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/269
resolves https://github.com/oncokb/oncokb-pipeline/issues/270


zhx828:
@benzuzu a bug needs to be fixed
- select the cancer type that in Inclusion which has been excluded
- delete the cancer type from exclusion
- the error message still is shown and confirm button is disabled which should be the reverse.

bprize15:
> @bprize15 a bug needs to be fixed
> 
> * select the cancer type that in Inclusion which has been excluded
> * delete the cancer type from exclusion
> * the error message still is shown and confirm button is disabled which should be the reverse.

Sorry about that, I should've caught it. It's fixed now.

zhx828 (approved):

Issue #209, opened 2024-02-27
Fix reference parsing in review mode diffs


bprize15:
In the diff library, if the change occurs within a PMID section (i. e `(PMID: 1)` -> `(PMID: 1, 2)`), the string is split up, and our regex that parses references can no longer parse the string.

It might be worth looking into creating our own diff viewer component or testing out other libraries.

PR #210, opened 2024-02-27
Add underline to references and link references in review mode

bprize15:
Related issue: https://github.com/oncokb/oncokb-transcript/issues/209
resolves https://github.com/oncokb/oncokb/issues/3691

### DIFF
--- a/src/main/webapp/app/config/constants/constants.ts
+++ b/src/main/webapp/app/config/constants/constants.ts
@@ -413,5 +413,6 @@ export const CHECKBOX_LABEL_LEFT_MARGIN = 18.75;
 export const FDA_SUBMISSION_REGEX = new RegExp('^([A-Z]+[0-9]+)(\\/((S[0-9]+)(-(S[0-9]+))?))?');
 export const UUID_REGEX = new RegExp('\\w{8}-\\w{4}-\\w{4}-\\w{4}-\\w{12}');
 export const INTEGER_REGEX = new RegExp('^\\d+$');
+export const REFERENCE_LINK_REGEX = /(\(.*?[PMID|NCT|Abstract].*?\))/i;
 
 export const CURATE_NEW_GENE_TEXT = 'Curate New Gene';
--- a/src/main/webapp/app/shared/links/TextWithRefs.tsx
+++ b/src/main/webapp/app/shared/links/TextWithRefs.tsx
@@ -1,0 +1,46 @@
+import { REFERENCE_LINK_REGEX } from 'app/config/constants/constants';
+import { ParsedRef, parseReferences } from 'app/oncokb-commons/components/RefComponent';
+import React from 'react';
+
+export interface ITextWithRefsProps {
+  content: string;
+}
+
+export default function TextWithRefs({ content }: ITextWithRefsProps) {
+  const result: JSX.Element[] = [];
+  const parts = content?.split(REFERENCE_LINK_REGEX) || [];
+
+  for (let i = 0; i < parts.length; i++) {
+    const part = parts[i];
+    if (REFERENCE_LINK_REGEX.exec(part)) {
+      const refs = parseReferences(part, true);
+      result.push(formatReferences(refs));
+    } else {
+      result.push(<span key={i}>{part}</span>);
+    }
+  }
+
+  return <span>{result}</span>;
+}
+
+function formatReferences(refs: ParsedRef[]) {
+  const prefix = refs[0].prefix;
+  const joinedReferences = refs.map((ref, index) => {
+    return (
+      <span key={index}>
+        <a style={{ textDecoration: 'underline' }} target="_blank" rel="noopener noreferrer" href={ref.link}>
+          {`${prefix}${ref.content}`}
+        </a>
+        {index !== refs.length - 1 ? ', ' : null}
+      </span>
+    );
+  });
+
+  return (
+    <span>
+      {'('}
+      {joinedReferences}
+      {')'}
+    </span>
+  );
+}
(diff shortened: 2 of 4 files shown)


src/main/webapp/app/shared/links/TextWithRefs.tsx
```diff
@@ -0,0 +1,47 @@
+import { REFERENCE_LINK } from 'app/config/constants/constants';
+import { ParsedRef, parseReferences } from 'app/oncokb-commons/components/RefComponent';
+import React from 'react';
+
+export interface ITexWithRefsProps {
+  content: string;
+}
+
+/* eslint-disable @typescript-eslint/prefer-regexp-exec */
```
zhx828, line 9: Do we need this?

src/main/webapp/app/shared/links/TextWithRefs.tsx
```diff
@@ -0,0 +1,47 @@
...
+export default function TextWithRefs({ content }: ITexWithRefsProps) {
+  const result: JSX.Element[] = [];
+  const parts = content?.split(REFERENCE_LINK) || [];
+
+  function formatReferences(refs: ParsedRef[]) {
```
zhx828, line 14: Can we move this function out of this component, so the var definition is closer to where it's actually used?

src/main/webapp/app/shared/links/TextWithRefs.tsx
```diff
@@ -0,0 +1,47 @@
...
+        {'('}
+        {joinedReferences}
+        {')'}
+      </span>
+    );
+  }
+
+  for (let i = 0; i < parts.length; i++) {
+    const part = parts[i];
+    if (part.match(REFERENCE_LINK)) {
```
zhx828, line 38: Will this ever be true? When you split the content using `REFERENCE_LINK`, wouldn't all ref been removed from parts?

zhx828 (changes requested):

src/main/webapp/app/shared/links/TextWithRefs.tsx
bprize15: Yes, since the component uses `Regex.match`

src/main/webapp/app/shared/links/TextWithRefs.tsx
bprize15: It will, since the abstract title is included in the capture group, it will be included in the output of `.split()`

src/main/webapp/app/shared/links/TextWithRefs.tsx
zhx828: Based on the check, wouldn't be using exec better?

src/main/webapp/app/shared/links/TextWithRefs.tsx
zhx828: The split will remove the abstract capture group though, would it?

bprize15: It will actually keep the abstract capture group. Here is an example: https://onecompiler.com/javascript/425tvvb94

src/main/webapp/app/shared/links/TextWithRefs.tsx
bprize15: I think either way works, so I changed it to `exec` so we don't need to disable eslint

src/main/webapp/app/shared/links/TextWithRefs.tsx
zhx828: Oh interesting. So if you have the regex capture group, it will remain in the list. https://onecompiler.com/javascript/425wyh6mq if you remove the capture group, it will remove the matched string.

zhx828 (approved):
LGTM, thanks

src/main/webapp/app/app.scss
```diff
@@ -7,6 +7,10 @@ html {
   font-size: $oncokb-font-font-size;
 }
 
+a {
+  text-decoration: underline;
+}
```
calvinlu3, line 12: This will give underline to all a tag on the page. For instance, the navigation sidebar uses <a> tags as part of the NavLinks I believe. I think we only need it for reference linkouts?

src/main/webapp/app/shared/links/TextWithRefs.tsx
```diff
@@ -0,0 +1,46 @@
+import { REFERENCE_LINK } from 'app/config/constants/constants';
+import { ParsedRef, parseReferences } from 'app/oncokb-commons/components/RefComponent';
+import React from 'react';
+
+export interface ITexWithRefsProps {
```
calvinlu3, line 5:
```suggestion
export interface ITextWithRefsProps {
```

src/main/webapp/app/config/constants/constants.ts
```diff
@@ -413,5 +413,6 @@ export const CHECKBOX_LABEL_LEFT_MARGIN = 18.75;
 export const FDA_SUBMISSION_REGEX = new RegExp('^([A-Z]+[0-9]+)(\\/((S[0-9]+)(-(S[0-9]+))?))?');
 export const UUID_REGEX = new RegExp('\\w{8}-\\w{4}-\\w{4}-\\w{4}-\\w{12}');
 export const INTEGER_REGEX = new RegExp('^\\d+$');
+export const REFERENCE_LINK = /(\(.*?[PMID|NCT|Abstract].*?\))/i;
```
calvinlu3, line 416:
```suggestion
export const REFERENCE_LINK_REGEX = /(\(.*?[PMID|NCT|Abstract].*?\))/i;
```

calvinlu3 (changes requested):

PR #213, opened 2024-02-29
Add search references

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3676

### DIFF
--- a/src/main/webapp/app/shared/util/utils.tsx
+++ b/src/main/webapp/app/shared/util/utils.tsx
@@ -12,6 +12,7 @@ import { IUser } from '../model/user.model';
 import { CancerType, Tumor } from '../model/firebase/firebase.model';
 import { ITreatment } from 'app/shared/model/treatment.model';
 import _ from 'lodash';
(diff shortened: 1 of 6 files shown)


src/main/webapp/app/shared/util/utils.tsx
```diff
@@ -308,3 +309,21 @@ export function extractPositionFromSingleNucleotideAlteration(alteration: string
     return null;
   }
 }
+
+/* eslint-disable @typescript-eslint/prefer-regexp-exec */
```
zhx828, line 313: I've seen this multiple time. Is it worthy for us just disable this check?

zhx828 (approved):

src/main/webapp/app/shared/util/utils.tsx
bprize15: Yeah I would say so. Will look into why we use that check in the first place, but it just feels inconvenient sometimes.

zhx828: It's the default from eslint. Feel free to fine tone it in https://github.com/oncokb/oncokb-transcript/blob/rc/.eslintrc.json

PR #214, opened 2024-03-01
Add iFrame to display Pubmed articles

calvinlu3:
This is related to https://github.com/oncokb/oncokb/issues/3675

The purpose is to test if showing pubmed article in an iFrame still works even if the header contains `X-Frame-Options: DENY`


zhx828:
We decided to pull the data from NIH directly. The code is added in https://github.com/oncokb/oncokb-transcript/pull/255

PR #216, opened 2024-03-01
Update collapsible color styles and add not curatable badge to mutation effect

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3679
This fixes https://github.com/oncokb/oncokb/issues/3677

Overall:
image

Mutation Effect Not Curatable:
image
image

### DIFF
--- a/src/main/webapp/app/config/colors.ts
+++ b/src/main/webapp/app/config/colors.ts
@@ -18,3 +18,10 @@ export const LEVELS = {
   R1: '#EE3424',
   R2: '#F79A92',
 };
+
+export const COLLAPSIBLE_LEVELS = {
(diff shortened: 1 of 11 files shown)


calvinlu3:
@zhx828 I think we probably need another round of feedback for the color changes. 
I think the following looks a bit strange:
1. The cancer type level background color is too similar to the textarea's light blue color. Maybe we need to darken the background colors for collapsibles?
2. This one probably won't matter when we get rid of germline. The white background for somatic/germline looks off.

calvinlu3:
image

zhx828:
@calvinlu3 I think it looks good and we should just go ahead implement and make it ready for review tomorrow.

PR #217, opened 2024-03-04
allow selecting ncit drug and add to drug list

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3678

### DIFF
--- a/src/main/webapp/app/shared/modal/ModifyTherapyModal.tsx
+++ b/src/main/webapp/app/shared/modal/ModifyTherapyModal.tsx
@@ -1,22 +1,25 @@
 import React, { useEffect, useMemo } from 'react';
 import { SimpleConfirmModal } from './SimpleConfirmModal';
-import { Treatment } from '../model/firebase/firebase.model';
 import DrugSelect, { DrugSelectOption } from '../select/DrugSelect';
 import { IRootStore } from 'app/stores';
 import { componentInject } from '../util/typed-inject';
 import { observer } from 'mobx-react';
 import { IDrug } from '../model/drug.model';
 import { FaExclamationCircle, FaRegTrashAlt } from 'react-icons/fa';
-import styles from './styles.module.scss';
 import './modify-therapy-modal.scss';
 import { Button } from 'reactstrap';
 import { generateUuid } from '../util/utils';
+import { parseNcitUniqId } from '../select/NcitCodeSelect';
+import _ from 'lodash';
+import { Treatment } from '../model/firebase/firebase.model';
+import { ERROR_EXCLAMATION_ICON_SIZE } from 'app/config/constants/constants';
 
 export interface IModifyTherapyModalProps extends StoreProps {
   treatmentUuid: string;
   treatmentName: string;
-  drugList: IDrug[];
-  onConfirm: (treatmentName: string) => void;
+  drugList: readonly IDrug[];
+  currentTreatments: Treatment[];
+  onConfirm: (treatmentName: string, newDrugs: IDrug[]) => void;
   onCancel: () => void;
 }
 
@@ -24,6 +27,7 @@ function ModifyTherapyModal({
   treatmentUuid,
   treatmentName,
   drugList,
+  currentTreatments,
   onConfirm,
   onCancel,
   modifyTherapyModalStore,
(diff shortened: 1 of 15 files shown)


src/main/webapp/app/shared/select/NcitCodeSelect.tsx
```diff
@@ -26,16 +26,17 @@ export const parseNcitUniqId = (ncitId: string) => {
   }
 };
 
+export function getOptionFromNcit(ncit: INciThesaurus) {
+  return {
```
zhx828, line 30: The following is the current drug option, we should use the same. We should also add ncit code in both places to support the feature.

```

          value: entity.id,
          label: entity.name,
          uuid: entity.uuid,
          synonyms: entity.nciThesaurus?.synonyms,
        }
```

zhx828 (changes requested):

zhx828:
Because the way the label is set up in https://github.com/oncokb/oncokb-transcript/pull/217#discussion_r1511502772, the language `The result will be added ...` is wrong. It should not include the code in the name.
Screenshot 2024-03-04 at 11 24 50 AM

src/main/webapp/app/shared/select/NcitCodeSelect.tsx
bprize15: Good call for adding synonyms. This will allow the drug select to support synonyms for NCITs. However, I don't think it makes sense for an NCIT to have a uuid. Since that is only relevant with firebase, and NCITs are not in firebase until the user hits `Confirm`, I think it makes sense to generate the uuid at that point. This is fine, since there is np reason the `ModifyTherapyModal` would try to access an NCIT's uuid.

src/main/webapp/app/shared/modal/ModifyTherapyModal.tsx
```diff
@@ -42,6 +44,7 @@ function ModifyTherapyModal({
   );
 }
 
+/* eslint-disable no-console */
```
zhx828, line 47: Should not be included.

src/main/webapp/app/shared/modal/ModifyTherapyModal.tsx
```diff
@@ -105,8 +110,28 @@ const ModifyTherapyModalContent = observer(
...
+          const newDrugs: IDrug[] = modifyTherapyModalStore.selectedTreatments.reduce((accumulator: IDrug[], currentTreatment) => {
+            const drugs = currentTreatment
+              .filter(therapy => therapy.ncit)
+              .map<IDrug>(therapy => ({
+                uuid: generateUuid(),
+                name: therapy.ncit.preferredName,
+                nciThesaurus: parseNcitUniqId(therapy.value),
+              }));
+            return accumulator.concat(drugs);
+          }, []);
```
zhx828, line 122: This newDrugs will have dups if I add the same drug in different treatments. Only one drug creation call will success as expected. But other ones will fail and the warning message will pop up.

zhx828 (changes requested):
- I noticed that you can add the same drug in multiple treatments which we should not allow.
- To keep the consistency, I think we should display the code for all drugs in the drug select for drugs that from Drug table and NCIT table. However, the hint  `The result will be added as ...` shouldn't include code because we don't display code typically in curation platform.

zhx828:
- You can actually add the same drug when modifying the cancer type. I think it has to do with the code in the name. If adding the code to the drug name when modifying, let's move the code to the second line with the synoms? 
Screenshot 2024-03-06 at 5 10 48 PM

- I think we talked about this, but we shouldn't allow to add the same treatment
Screenshot 2024-03-06 at 5 07 59 PM

src/main/webapp/app/shared/modal/ModifyTherapyModal.tsx
```diff
@@ -43,9 +48,76 @@ function ModifyTherapyModal({
...
+    treatmentName,
+    drugList,
+    currentTreatments,
+    onConfirm,
+    onCancel,
+    modifyTherapyModalStore,
+  }: IModifyTherapyModalProps) => {
     const disableDeleteTherapy = modifyTherapyModalStore.selectedTreatments.length < 2;
     const isEmptyTherapy = modifyTherapyModalStore.selectedTreatments.some(therapy => therapy.length === 0);
+    const isDuplicate = isDuplicateTherapy();
```
zhx828, line 62: What's the point of adding this when you already have alreadyExists?

bprize15: `isDuplicate` refers to when there are multiple instances of the same treatment within the modal (i.e. to dropdowns contain exactly the same drugs). `alreadyExists` refers to when the therapy as a whole is already present for the cancer type (i.e. there is already a collapsible corresponding to that therapy). I have changed the name of `isDuplicateTherapy` to `isDuplicateTreatment` to clear this up.

PR #219, opened 2024-03-04
add sorting to mutation list

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3695


zhx828:
- You can actually add the same drug when modifying the cancer type. I think it has to do with the code in the name. If adding the code to the drug name when modifying, let's move the code to the second line with the synoms? 
Screenshot 2024-03-06 at 5 10 48 PM

- I think we talked about this, but we shouldn't allow to add the same treatment
Screenshot 2024-03-06 at 5 07 59 PM

src/main/webapp/app/shared/util/firebase/firebase-utils.ts
```diff
@@ -318,3 +334,136 @@ export const isMutationEffectCuratable = (mutationName: string) => {
...
+      return order;
+    }
+
+    index++;
+  }
+
+  return 0;
+}
+
+export function compareMutationsByOncogenicity(mut1: Mutation, mut2: Mutation) {
```
zhx828, line 374: Visually since we gonna use the same blue icon for oncogenic/likely/resistance, it's a bit confusing sorting based on their priority then by position. Let's try the priority of oncogenic/likely/resistance the same.

zhx828 (approved):
LGTM, thanks.

PR #220, opened 2024-03-07
Resize sidebar

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/279#issue-2172097616

### DIFF
--- a/src/main/webapp/app/components/sidebar/OncoKBSidebar.tsx
+++ b/src/main/webapp/app/components/sidebar/OncoKBSidebar.tsx
@@ -1,49 +1,174 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import './oncokb-sidebar.scss';
 import { observer } from 'mobx-react';
 import { IRootStore } from 'app/stores';
 import { componentInject } from 'app/shared/util/typed-inject';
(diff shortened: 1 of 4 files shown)


src/main/webapp/app/components/sidebar/OncoKBSidebar.tsx
```diff
@@ -15,27 +18,144 @@ const OncoKBSidebar = ({
...
+      }
+    }
+
+    function onMouseUp() {
+      setHoldingSidebarBorder(false);
+      document.body.removeEventListener('mousemove', onMouseMove);
+    }
+
+    document.body.addEventListener('mousemove', onMouseMove);
+    document.body.addEventListener('mouseup', onMouseUp, { once: true });
```
calvinlu3, line 92: Looks like the `mouseup` listener is not removed? Maybe, the inner functions can be moved one level out and then the useEffect can remove listener when unmounted?

calvinlu3:
I wonder if its a good idea to also show another collapse icon (chevron double) to the minWidth when the sidebar is currently at its maxWidth.
image

Just a thought. I'm fine with the current design.

calvinlu3:
Also, the collapse/expand icons using the circle seems a bit odd. The button style probably looks better?

PR #222, opened 2024-03-07
Fix VUS column width

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3699


calvinlu3:
@bprize15 Just a one line change to make sure the badge is on the same line as timestamp. Fine when using 14' screen

bprize15 (approved):
LGTM

PR #225, opened 2024-03-12
Review mockups

calvinlu3:
This fixes https://github.com/oncokb/oncokb/issues/3698

### DIFF
--- a/src/main/webapp/app/pages/curation/collapsible/Collapsible.tsx
+++ b/src/main/webapp/app/pages/curation/collapsible/Collapsible.tsx
@@ -21,6 +21,7 @@ export interface CollapsibleProps {
   title: React.ReactNode;
   disableLeftBorder?: boolean;
   borderLeftColor?: string;
+  backgroundColor?: string;
   info?: React.ReactNode;
   action?: React.ReactNode;
   children: React.ReactNode;
@@ -45,6 +46,7 @@ export default function Collapsible({
   isSectionEmpty = false,
   disableCollapsible = false,
   isPendingDelete = false,
+  backgroundColor,
   badgeOverride,
   onToggle,
 }: CollapsibleProps) {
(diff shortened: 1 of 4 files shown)


src/main/webapp/app/pages/curation/collapsible/Collapsible.tsx
```diff
@@ -83,14 +85,23 @@ export default function Collapsible({
         )}
         ref={node => {
           if (node && borderLeftColor) {
+            const hadBackgroundColor = backgroundColor;
             if (disableCollapsible) {
               borderLeftColor = DISABLED_NEST_LEVEL_COLOR;
+              if (!hadBackgroundColor) {
+                backgroundColor = DISABLED_NEST_LEVEL_COLOR;
```
bprize15, line 92: Do we need `hadBackgroundColor`? Can we just change to `if (!backgroundColor)`?

src/main/webapp/app/shared/util/firebase/firebase-review-utils.ts
```diff
@@ -352,6 +355,9 @@ const findMutationEffectReviews = (
         { [fieldKey]: mutationEffect[reviewKey].lastReviewed },
         isNestedUnderCreateOrDelete(parentReview)
       );
+      if (fieldKey === 'oncogenic' || fieldKey === 'effect') {
```
bprize15, line 358: Solution works for now, but will break with new listener changes. Can leave as is, but might want to consider a fix that checks the input type or some way that will not break (if that is possible).

src/main/webapp/app/pages/curation/collapsible/Collapsible.tsx
calvinlu3: Mainly used in:
```
getHexColorWithAlpha(backgroundColor, hadBackgroundColor ? 0 : 0.05)
```
By default, the background color = border left color with 0.05 opacity. But if background color was passed as prop, then it should not have opacity.

src/main/webapp/app/shared/util/firebase/firebase-review-utils.ts
calvinlu3: Yeah this is true. Don't know if there's a good way to know if a field is a checkbox/radio without probably hard-coding. I will come back and fix it

bprize15 (approved):

PR #228, opened 2024-03-12
Feat/refactor listeners

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/285


src/main/webapp/app/shared/icons/CommentIcon.tsx
```diff
@@ -58,6 +67,45 @@ const CommentIcon = observer((props: ICommentIconProps) => {
     timeoutId.current = id;
   }
 
+  async function handleCreateComment(content: string) {
+    // replace with runTransaction?
```
zhx828, line 71: runTransaction is meaningful when there are multiple data elements are involved. I don't think you need one for this method.

src/main/webapp/app/pages/curation/nestLevelSummary/MutationLevelSummary.tsx
```diff
@@ -3,25 +3,36 @@ import { IRootStore } from 'app/stores';
...
-  const mutationLevelSummary = mutationSummaryStats[mutationUuid];
-  return <NestLevelSummary summaryStats={mutationLevelSummary} hideOncogenicity={hideOncogenicity} />;
+const MutationLevelSummary = ({ mutationPath, firebaseDb, hideOncogenicity = false }: MutationLevelSummaryProps) => {
+  const [mutationStats, setMutationStats] = useState(undefined);
+
+  useEffect(() => {
+    const callbacks = [];
+    callbacks.push(
+      onValue(ref(firebaseDb, mutationPath), snapshot => {
+        const calcMutationStats = getMutationStats(snapshot.val());
```
zhx828, line 23: Thoughts on creating a delayed calculation, like debounce here?

zhx828: Same can be applied to other **Summary

src/main/webapp/app/pages/curation/mutation/MutationsFilterSection.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+  const [mutationEffectFilter, setMutationEffectFilter] = useState(initFilterCheckboxState(MUTATION_EFFECT_OPTIONS));
+  const [tempMutationEffectFilter, setTempMutationEffectFilter] = useState(initFilterCheckboxState(MUTATION_EFFECT_OPTIONS));
+
+  const [txLevelFilter, setTxLevelFilter] = useState(initFilterCheckboxState(TX_LEVEL_OPTIONS));
+  const [tempTxLevelFilter, setTempTxLevelFilter] = useState(initFilterCheckboxState(TX_LEVEL_OPTIONS));
+
+  useEffect(() => {
+    const unsubscribe = onValue(ref(firebaseDb, mutationsPath), snapshot => {
+      setMutations(snapshot.val());
```
zhx828, line 37: also here adding a debounce seems very worthy.

zhx828 (changes requested):
Overall looks pretty awesome. A lot of components have been tidied up. Especially glad to see MC is only 200 lines of code from 800! Thank you @bprize15 @calvinlu3 

-  [x] all VUS items seems to use the same timestamp
Screenshot 2024-03-18 at 10 03 24 AM
- [ ] not sure this is new, but changing cancer type exclusion would not trigger a review, but it should
- [ ] is there a way for us to monitor how many listeners are attached to the firebase in total?

calvinlu3:
@zhx828 
```
not sure this is new, but changing cancer type exclusion would not trigger a review, but it should
```

This issue is tracked by https://github.com/oncokb/oncokb-pipeline/issues/267. Issue is basically that all name changes are not tracked right now, so will not show up in review mode.

src/main/webapp/app/pages/curation/list/FirebaseList.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  useEffect(() => {
+    const listRef = ref(firebaseDb, path);
+    const unsubscribe = onValue(listRef, snapshot => {
+      if (!snapshot.val()) {
+        // this happens for emtpy array, such as tumors for a new mutation
+        return;
+      }
+
+      if (!indices) {
```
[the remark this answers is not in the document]
bprize15: You are exactly correct, we don't need to update indices given that assumption, and that simplifies the logic.

We also know that a curator cannot edit the gene while another is reviewing the gene (Calvin mentioned that this is the intended behavior but not implemented yet), so the assumption should remain correct in all cases.

src/main/webapp/app/pages/curation/nestLevelSummary/MutationLevelSummary.tsx
bprize15: I agree, will save some work and not recalculate every keystroke.

PR #243, opened 2024-03-19
Update therapy dropdowns

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/293

The dropdown selections were fixed in the refactoring listeners PR. In this PR, we don't show liquid/solid tumor levels of evidence when highest level is resistance level.

Also contains bug fixes:
1. Remove collapsible after accepting changes
2. Don't show info/action components for collapsibles that are pending delete


zhx828 (approved):
Looks good. Thanks.

PR #252, opened 2024-03-26
Add button mutation modal

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/302

### DIFF
--- a/src/main/webapp/app/shared/modal/AddVusModal.tsx
+++ b/src/main/webapp/app/shared/modal/AddVusModal.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import CreatableSelect from 'react-select/creatable';
 import { Mutation, VusObjList } from '../model/firebase/firebase.model';
 import { parseAlterationName } from '../util/utils';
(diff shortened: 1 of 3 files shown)


src/main/webapp/app/shared/modal/AddVusModal.tsx
```diff
@@ -54,12 +55,16 @@ const AddVusModalV2 = (props: IAddVusModalProps) => {
...
 
+  function handleVariantAdded() {
+    const filteredAlterations = filterAlterationsAndNotify(inputValue);
+    setVariants(state => [...state, ...filteredAlterations.map(alt => createOption(alt))]);
+    setInputValue('');
+  }
+
   const handleKeyDown = event => {
     if (!inputValue) return;
     if (event.key === 'Enter' || event.key === 'tab') {
```
calvinlu3, line 66: I think we don't need to check for tab anymore?

calvinlu3 (approved):
LGTM. The excluded cancer type select still has text saying "Enter alteration(s) and press enter"

src/main/webapp/app/shared/modal/AddVusModal.tsx
bprize15: I think it's fine. I don't think there's any harm in allowing the tab key to behave like the enter key, I see online that some people expect that. Will change the language though as you mentioned.

PR #262, opened 2024-03-29
Add validation tool

jfkonecn:
resolves oncokb/oncokb#3703


src/main/java/org/mskcc/oncokb/curation/web/rest/ClientForwardController.java
```diff
@@ -40,7 +40,7 @@ public ResponseEntity<String> index() {
      * Forwards any unmapped paths (except those containing a period) to the client {@code index.html}.
      * @return forward to client {@code swagger-ui/index.html}.
      */
-    @GetMapping(value = "/**/{path:[^\\.]*}")
+    // @GetMapping(value = "/**/{path:[^\\.]*}")
     public String forward() {
```
jfkonecn, line 44: Need to fix this. It's blocking websocket from working.

zhx828:
Some feature requests before code review
- for success test/info, we should not allow the collapsible to open. See deleted `test` mutation in https://curation.oncokb.org/curation/ZNRF3/somatic as an example
- we need to notify the user the validation is done. Maybe add a toast banner?
- Let's change the Variant column header to `Biomarker`
- the Validate button does not need to be wrapped in a container again. Otherwise you will see extra white space on the left of the button comparing to Tools tab
- I don't think the refresh icon should be shown up before the first time of validation. And the button should be primary button before first time of validation. See `Release Gene` button in the gene page under the Tools tab

src/main/webapp/app/components/tabs/CurationDataValidationTab.tsx
```diff
@@ -0,0 +1,330 @@
...
+    error: true,
+  });
+  useEffect(() => {
+    onChange(validationResultFilter);
+  }, [validationResultFilter]);
+
+  return (
+    <Form className={className}>
+      <Row>
+        <Col>
```
bprize15, line 34: Could you remove the padding on the column so it lines up nicely with the tab titles?

src/main/webapp/app/components/tabs/CurationDataValidationTab.tsx
```diff
@@ -0,0 +1,330 @@
...
+  const [isValidating, setIsValidating] = React.useState(false);
+  const [validationCounter, setValidationCounter] = React.useState(0);
+
+  useEffect(() => {
+    setIsValidating(validationCounter > 0);
+  }, [validationCounter]);
+
+  const onClose = useCallback(() => {
+    setIsValidating(false);
+    toast.success('Validation completed!');
```
bprize15, line 239: We have utility functions `notifySuccess` and `notifyError` for this

src/main/webapp/app/components/tabs/CurationDataValidationTab.tsx
```diff
@@ -0,0 +1,330 @@
...
+  }, []);
+  const onOpen = useCallback(() => {
+    return;
+  }, []);
+  const onError = useCallback(() => {
+    toast.error('Error while attempting to validate. Please try again.');
```
bprize15, line 245: Can use `notifyError` here

src/main/webapp/app/pages/curation/collapsible/Collapsible.tsx
```diff
@@ -25,17 +34,41 @@ export default function Collapsible({
...
   const displayOptionsOverride = useMemo(() => {
+    memoRerenderRef.current++;
     if (isPendingDelete) {
       defaultDisplayOptions.disableCollapsible = true;
       defaultDisplayOptions.hideAction = displayOptions?.hideAction || false;
       defaultDisplayOptions.hideInfo = displayOptions?.hideInfo || false;
     }
+    const displayOptionsJson = JSON.stringify(displayOptions);
+    const oldDisplayOptionsJson = JSON.stringify(displayOptionRef.current);
+    if (isPendingDeleteRef.current === isPendingDelete && displayOptionsJson === oldDisplayOptionsJson) {
```
bprize15, line 52: Can delete this I'd imagine

.nvmrc
```diff
@@ -0,0 +1 @@
+14.17.6
```
zhx828, line 1: I noticed this is added in few of your prs. When .nvmrc file is specified, nvm will switch to the version automatically?

src/main/docker/keycloak.yml
```diff
@@ -1,30 +1,40 @@
...
+      POSTGRES_DB: keycloak
+      POSTGRES_USER: root
+      POSTGRES_PASSWORD: root
     volumes:
-      - ./realm-config:/opt/jboss/keycloak/realm-config
-      # Keycloak allows you to create custom themes: https://www.keycloak.org/docs/latest/server_development/#_themes
-      # - ./realm-config/themes/custom-theme:/opt/jboss/keycloak/themes/custom-theme
+      - kc-db:/var/lib/mysql
+    networks:
+      - kc-network
```
zhx828, line 14: Curious why we create a new network

src/main/java/org/mskcc/oncokb/curation/web/websocket/ProxyCurationValidationApiHandler.java
```diff
@@ -0,0 +1,55 @@
...
+public class ProxyCurationValidationApiHandler implements WebSocketHandler {
+
+    private OncoCoreWebSocketHandler oncoHandler;
+    private String validationEndpoint;
+
+    public ProxyCurationValidationApiHandler(ApplicationProperties applicationProperties) {
+        String url = applicationProperties.getOncokb().getUrl();
+        url = url.replace("https://", "wss://");
+        url = url.replace("http://", "ws://");
+        this.validationEndpoint = url + "/api/websocket/curation/validation";
```
zhx828, line 22: Should we move this to the application as well?

src/main/webapp/app/components/tabs/CurationDataValidationTab.tsx
```diff
@@ -0,0 +1,330 @@
...
+    setIsValidating(false);
+    notifySuccess('Validation completed!');
+  }, []);
+  const onOpen = useCallback(() => {
+    return;
+  }, []);
+  const onError = useCallback(() => {
+    notifyError(new Error('Error while attempting to validate. Please try again.'));
+  }, []);
+  const url = relativeWebsocketUrlToAbsolute('/api/websocket/curation/validation');
```
zhx828, line 247: I would prefer creating a websocket API in curation platform and let the server side routes certain URL pattern to the core service so we could also add other websockets as well.

zhx828 (changes requested):
For some reason, the validation websocket doesn't work after recent changes. I will test again.

zhx828:
Also thanks for testing useMemo. I will take another look at it when possible. What's the conclusion on that?

jfkonecn:
> Also thanks for testing useMemo. I will take another look at it when possible. What's the conclusion on that?

It rerenders every time.
According to the [react docs](https://react.dev/reference/react/useMemo) it uses [Object.is()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) to do the equality check which does a pointer check for the equality between two objects.

zhx828 (approved):
Could you create a ticket to follow through on swagger/util as we discussed? Thanks!

PR #264, opened 2024-04-01
Convert between VUS and mutation 

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/257

### DIFF
--- a/src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
+++ b/src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
@@ -15,12 +15,18 @@ import EditIcon from 'app/shared/icons/EditIcon';
 import AddMutationModal from 'app/shared/modal/AddMutationModal';
 import ModifyCancerTypeModal from 'app/shared/modal/ModifyCancerTypeModal';
 import { Alteration, Review } from 'app/shared/model/firebase/firebase.model';
-import { getMutationName, isMutationEffectCuratable, isSectionRemovableWithoutReview } from 'app/shared/util/firebase/firebase-utils';
+import {
+  getFirebaseGenePath,
+  getFirebaseVusPath,
+  getMutationName,
+  isMutationEffectCuratable,
+  isSectionRemovableWithoutReview,
+} from 'app/shared/util/firebase/firebase-utils';
 import { componentInject } from 'app/shared/util/typed-inject';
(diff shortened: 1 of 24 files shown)


src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
```diff
@@ -379,6 +420,16 @@ const MutationCollapsible = ({
...
         />
       ) : undefined}
+      {isConvertingToVus ? (
+        <AddVusModal
+          hugoSymbol={hugoSymbol}
+          isGermline={isGermline}
+          vusList={vusData}
+          onCancel={() => setIsConvertingToVus(false)}
+          onConfirm={handleDemoteToVus}
+          convertOptions={{ initialAlterations: mutationName.split(', '), isConverting: true, mutationUuid }}
```
zhx828, line 430: Maybe true for newly added, but not true for existing string mutation
```suggestion
          convertOptions={{ initialAlterations: mutationName.split(','), isConverting: true, mutationUuid }}
```

src/main/webapp/app/shared/icons/MutationConvertIcon.tsx
```diff
@@ -0,0 +1,84 @@
...
+  ...actionIconProps
+}: IMutationConvertIconProps) => {
+  const [mutationList, setMutationList] = useState<Mutation[]>(null);
+  const [isDisabled, setIsDisabled] = useState(false);
+
+  useEffect(() => {
+    let unsubscribe;
+    if (convertTo === 'vus') {
+      unsubscribe = onValue(ref(firebaseDb, firebaseMutationsPath), snapshot => {
+        setMutationList(snapshot.val());
```
zhx828, line 37: This is that expensive listener again. Is it really necessary to listen on the whole mutations?

src/main/webapp/app/shared/icons/MutationConvertIcon.tsx
```diff
@@ -0,0 +1,84 @@
...
+      });
+    }
+    return () => unsubscribe?.();
+  }, []);
+
+  useEffect(() => {
+    if (mutationList) {
+      const newMutationName = parseAlterationName(mutationName)[0].alteration.split(', ');
```
zhx828, line 45: We should always use `,` instead

zhx828 (approved):
Some minor comments, overall looks good to me.

PR #265, opened 2024-04-02
fix no entry badge showing up incorrectly on mutation level

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/314


calvinlu3 (approved):
LGTM. Seems like the issue was that the mutation level stats did not calculate the tx stats.

PR #267, opened 2024-04-03
Fix mutation modal confirm

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/316

### DIFF
--- a/src/main/webapp/app/shared/modal/AddMutationModal.tsx
+++ b/src/main/webapp/app/shared/modal/AddMutationModal.tsx
@@ -1030,13 +1030,11 @@ interface IAddMutationModalFieldProps {
 function AddMutationModalField({ label, value: value, placeholder, onChange, isLoading }: IAddMutationModalFieldProps) {
   return (
     <Row className="align-items-center mb-3">
-      <Col className="px-0 col-3 mr-3">
-        <span className="mr-2">{label}</span>
-        {isLoading && (
-          <span>
-            <Spinner size="sm" />
-          </span>
-        )}
+      <Col className="px-0 col-3 mr-3 align-items-center">
+        <div className="d-flex align-items-center">
+          <span className="mr-2">{label}</span>
+          {isLoading && <Spinner color="primary" size="sm" />}
+        </div>
       </Col>
       <Col className="px-0">
         <Input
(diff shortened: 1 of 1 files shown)


src/main/webapp/app/shared/modal/AddMutationModal.tsx
```diff
@@ -359,8 +361,9 @@ function AddMutationModal({
   }
 
   const fetchNormalAlterationDebounced = useCallback(
-    _.debounce((newAlteration: string, alterationIndex: number, alterationData: AlterationData[]) => {
-      fetchNormalAlteration(newAlteration, alterationIndex, alterationData);
+    _.debounce(async (newAlteration: string, alterationIndex: number, alterationData: AlterationData[]) => {
+      await fetchNormalAlteration(newAlteration, alterationIndex, alterationData);
```
calvinlu3, line 365: We should probably handle scenario when the async function fails. I can create a ticket to make sure that all async function errors are caught.

src/main/webapp/app/shared/modal/AddMutationModal.tsx
```diff
@@ -1009,13 +1024,19 @@ interface IAddMutationModalFieldProps {
...
   return (
     <Row className="align-items-center mb-3">
       <Col className="px-0 col-3 mr-3">
-        <span>{label}</span>
+        <span className="mr-2">{label}</span>
+        {isLoading && (
+          <span>
+            <Spinner size="sm" />
+          </span>
+        )}
```
calvinlu3, line 1039: Could just be me, but the spinner position seems slightly below the center.

calvinlu3: Also, what do you think about using $oncokb-blue for spinner color (same as LoadingIndicator).

calvinlu3: image

src/main/webapp/app/shared/modal/AddMutationModal.tsx
bprize15: Ok sounds good, we can come back through for more error checking.

calvinlu3 (approved):
LGTM

PR #268, opened 2024-04-05
 Show hotspot icon and exon info on mutation

zhx828:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/283 and https://github.com/oncokb/oncokb/issues/3694

### DIFF
--- a/src/main/java/org/mskcc/oncokb/curation/domain/enumeration/HotspotFlagEnum.java
+++ b/src/main/java/org/mskcc/oncokb/curation/domain/enumeration/HotspotFlagEnum.java
@@ -1,0 +1,9 @@
+package org.mskcc.oncokb.curation.domain.enumeration;
+
+/**
+ * The TranscriptFlagEnum enumeration.
+ */
+public enum HotspotFlagEnum {
+    HOTSPOT_V1,
+    THREE_D,
+}
--- a/src/main/java/org/mskcc/oncokb/curation/repository/AlterationRepository.java
+++ b/src/main/java/org/mskcc/oncokb/curation/repository/AlterationRepository.java
@@ -57,7 +57,7 @@ public interface AlterationRepository extends JpaRepository<Alteration, Long>, J
         " left join fetch alteration.genes g" +
         " left join fetch alteration.flags" +
         " where g.id in :genes and " +
-        " alteration.consequence = :consequence and ((alteration.start <= :end and alteration.end >= :start) or (alteration.end >= :start and alteration.start <= :end)) "
+        " alteration.consequence = :consequence and (alteration.start <= :end and alteration.end >= :start)"
     )
     List<Alteration> findByGeneAndConsequenceThatOverlap(
         @Param("genes") List<Long> genes,
--- a/src/main/java/org/mskcc/oncokb/curation/repository/AlterationRepository.java
+++ b/src/main/java/org/mskcc/oncokb/curation/repository/AlterationRepository.java
@@ -2,8 +2,11 @@ package org.mskcc.oncokb.curation.repository;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import org.javers.spring.annotation.JaversSpringDataAuditable;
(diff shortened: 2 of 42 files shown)


src/main/java/org/mskcc/oncokb/curation/repository/AlterationRepository.java
```diff
@@ -42,12 +45,27 @@ public interface AlterationRepository extends JpaRepository<Alteration, Long>, J
...
         " (alteration.name =:query or alteration.alteration=:query )"
     )
     List<Alteration> findByNameOrAlterationAndGenesId(@Param("query") String query, @Param("id") Long geneId);
 
+    @Query(
+        "select distinct alteration from Alteration alteration" +
+        " left join fetch alteration.genes g" +
+        " left join fetch alteration.flags" +
+        " where g.id in :genes and " +
+        " alteration.consequence = :consequence and ((alteration.start <= :end and alteration.end >= :start) or (alteration.end >= :start and alteration.start <= :end)) "
```
calvinlu3, line 60:
```suggestion
        " alteration.consequence = :consequence and (alteration.start <= :end and alteration.end >= :start) "
```

This should be sufficient to check if two ranges overlap. The original had two duplicated conditioned with OR.

calvinlu3 (changes requested):

src/main/java/org/mskcc/oncokb/curation/domain/enumeration/HotspotFlagEnum.java
```diff
@@ -0,0 +1,9 @@
+package org.mskcc.oncokb.curation.domain.enumeration;
+
+/**
+ * The TranscriptFlagEnum enumeration.
+ */
+public enum HotspotFlagEnum {
+    HOTSPOT_V1,
```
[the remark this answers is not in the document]
zhx828: @calvinlu3 there is a V2 about to come out. I'm preping for that.

src/main/webapp/app/pages/curation/mutation/MutationsSection.tsx
```diff
@@ -12,25 +12,47 @@ import { componentInject } from 'app/shared/util/typed-inject';
...
   const [openMutationCollapsibleIndex, setOpenMutationCollapsibleIndex] = useState<number>(null);
 
   const mutationSectionRef = useRef<HTMLDivElement>(null);
   const mutationScrollContainerRef = useRef<HTMLDivElement>(null);
 
+  useEffect(() => {
+    onValue(
+      ref(firebaseDb, `${getFirebaseGenePath(isGermline, hugoSymbol)}/mutations`),
+      snapshot => {
+        annotatedAltsCache.fetch(hugoSymbol, snapshot.val());
```
calvinlu3, line 50: `fetch` takes `hugoSymbol` and `MutationQuery[]`, but `snapshot.val()` here is actually of type `Mutation[]`.

src/main/webapp/app/pages/curation/mutation/MutationsSection.tsx
```diff
@@ -12,25 +12,55 @@ import { componentInject } from 'app/shared/util/typed-inject';
...
+            return {
+              name: mut.name,
+              alterations: mut.alterations,
+            };
+          })
+        );
+      },
+      { onlyOnce: true }
+    );
+  }, []);
```
bprize15, line 62: Probably safer to add dependencies to dependency array

src/main/webapp/app/shared/links/Linkout.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { CSSProperties } from 'react';
+
+export const Linkout: React.FunctionComponent<{
+  to: string;
+  className?: string;
+  style?: CSSProperties;
+  addHttpsProtocol?: boolean;
+}> = props => {
+  let updatedLink = props.to;
+  if (props.addHttpsProtocol === undefined || props.addHttpsProtocol) {
```
bprize15, line 10: Are we defaulting to true here? In that case, think we should use default assignment

PR #275, opened 2024-04-12
Added sentry

jfkonecn:
resolves oncokb/oncokb-pipeline#325

### DIFF
--- a/src/main/resources/config/application-dev.yml
+++ b/src/main/resources/config/application-dev.yml
@@ -105,6 +105,18 @@ jhipster:
         # Token is valid 24 hours
         token-validity-in-seconds: 86400
         token-validity-in-seconds-for-remember-me: 2592000
+
+# ===================================================================
+# Sentry specific properties
+# How to configure Sentry Spring Boot: https://docs.sentry.io/clients/java/config/
+# Full reference is available at: https://github.com/getsentry/sentry-java/blob/main/sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryProperties.java
+# ===================================================================
+sentry:
+  dsn:
+  stacktrace:
+    app:
+      packages: org.mskcc.oncokb.curation;
+
 # ===================================================================
 # Application specific properties
 # Add your own application properties here, see the ApplicationProperties class
--- a/src/main/resources/config/application-dev.yml
+++ b/src/main/resources/config/application-dev.yml
@@ -150,3 +150,5 @@ application:
     measurement-id:
     service-account-credentials-path:
   oncokb-data-repo-dir:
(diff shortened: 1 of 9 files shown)


src/main/webapp/app/appConfig.ts
```diff
@@ -1,5 +1,6 @@
 type FrontendProperties = {
   firebase: FirebaseProperties;
+  sentryDsn: string;
```
calvinlu3, line 3: Could you help make the FrontendProperties optional and also `firebase` and `sentryDsn` optional in the type. We may have a scenario where we don't enable Firebase or Sentry or both. In that case, these properties can be undefined.

src/main/webapp/app/index.tsx
```diff
@@ -8,6 +8,74 @@ import setupAxiosInterceptors from './config/axios-interceptor';
 import ErrorBoundary from './shared/error/error-boundary';
 import AppComponent from './app';
 import { loadIcons } from './config/icon-loader';
+import { AppConfig } from './appConfig';
+import * as Sentry from '@sentry/react';
+
+if (AppConfig.serverConfig?.frontend.sentryDsn) {
```
calvinlu3, line 14: After we change the frontend properties to optional, we should also add `?.sentryDsn`

src/main/resources/config/application-dev.yml
```diff
@@ -105,6 +105,18 @@ jhipster:
...
+# Full reference is available at: https://github.com/getsentry/sentry-java/blob/main/sentry-spring-boot-starter/src/main/java/io/sentry/spring/boot/SentryProperties.java
+# ===================================================================
+sentry:
+  dsn:
+  stacktrace:
+    app:
+      packages: org.mskcc.oncokb.curation;
+
 # ===================================================================
 # Application specific properties
```
calvinlu3, line 121: Under application specific properties, could we add
```
frontend:
  sentryDsn: 
```

calvinlu3:
Also, could you update oncokb-deployment repo's configmaps: `oncokb_curation.yaml` and `oncokb_curation_beta.yaml` with the Sentry properties.

https://github.com/knowledgesystems/oncokb-deployment/blob/master/credentails/configmaps/oncokb-curation-beta.yaml

src/main/resources/config/application-dev.yml
jfkonecn: Do you mind if I do sentry-dsn since it looks like we're using dashes for space for everything else?

calvinlu3: Yes, please use the dash.

calvinlu3 (approved):

PR #278, opened 2024-04-17
Refactor history

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/326

### DIFF
--- a/src/main/webapp/app/config/constants/firebase.ts
+++ b/src/main/webapp/app/config/constants/firebase.ts
@@ -312,7 +312,7 @@ export enum READABLE_FIELD {
   MUTATION_EFFECT = 'Mutation Effect',
   PATHOGENIC = 'Pathogenic',
   ONCOGENIC = 'Oncogenic',
-  EFFECT = 'Effect',
+  EFFECT = '',
   MUTATION_SPECIFIC_PENETRANCE = 'Mutation Specific Penetrance',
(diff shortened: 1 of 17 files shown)


src/main/webapp/app/shared/util/firebase/firebase-history-utils.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+                cancerTypes_uuid: '456',
+              },
+            ],
+          },
+        ],
+        name: 'V600E',
+        name_review: {
+          added: true,
+          updateTime: 1550458712327,
+          updatedBy: 'Lindsay LaFave',
```
calvinlu3, line 67:
```suggestion
          updatedBy: 'User',
```

src/main/webapp/app/config/constants/firebase.ts
```diff
@@ -312,7 +312,7 @@ export enum READABLE_FIELD {
   MUTATION_EFFECT = 'Mutation Effect',
   PATHOGENIC = 'Pathogenic',
   ONCOGENIC = 'Oncogenic',
-  EFFECT = 'Effect',
+  EFFECT = '',
```
calvinlu3, line 315: Could we use `Effect` here?
Use case would be in review mode:
image

bprize15: Yeah for sure

calvinlu3 (approved):

PR #279, opened 2024-04-17
make mutation type labels readable

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3705


calvinlu3 (commented):
Don't think this is in a rush to get out. Let's hold off and ask Hongxin about the changes.

zhx828:
@calvinlu3 @bprize15 thanks. It looks good to me.

PR #280, opened 2024-04-22
add mutation store and convert icon store

bprize15:
resolves https://github.com/oncokb/oncokb/issues/3730


calvinlu3 (approved):
Works pretty well locally.

PR #283, opened 2024-04-24
Added api proxy to core

jfkonecn:
resolves oncokb/oncokb-pipeline#336


src/main/java/org/mskcc/oncokb/curation/web/rest/ApiProxy.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.http.converter.StringHttpMessageConverter;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.RestTemplate;
+import org.springframework.web.server.ResponseStatusException;
+
+@RestController
+@RequestMapping({ "/api", "/legacy-api" })
```
zhx828, line 23: proxy all `/api` as well?

src/main/java/org/mskcc/oncokb/curation/service/ApiProxyService.java
```diff
@@ -0,0 +1,52 @@
...
+    public HttpHeaders prepareHttpHeaders(String contentType) {
+        HttpHeaders httpHeaders = new HttpHeaders();
+        if (contentType != null) {
+            httpHeaders.setContentType(MediaType.valueOf(contentType));
+        }
+
+        return httpHeaders;
+    }
+
+    public ResponseEntity<String> exchangeRequest(String body, HttpMethod method, String contentType, String apiRequest)
```
zhx828, line 44: This method doesn't seem to be used.

zhx828 (changes requested):
Should we get rid of the following files from the shared/api/generated folder?

- .gitignore
- .npmignore
- .openapi-generator
- .openapi-generator-ignore

src/main/java/org/mskcc/oncokb/curation/web/rest/ApiProxy.java
jfkonecn: I thought we wanted all of core's API's. Do we just want the /legacy-api?

src/main/java/org/mskcc/oncokb/curation/service/ApiProxyService.java
jfkonecn: I may have done some copy and pasting....
I'll take care of it.

.prettierrc
```diff
@@ -1,5 +1,8 @@
 # Prettier configuration
 
+plugins:
+  - prettier-plugin-java
```
jfkonecn, line 4: @zhx828 I was having this [issue](https://github.com/jhipster/prettier-java/issues/508) where prettier couldn't figure out which parser to use for java files so I forced it to use the java parser.

zhx828 (approved):
Thanks! Don't forget to squash before merging.

zhx828:
@jfkonecn the CI test failed.

PR #284, opened 2024-04-25
Refactor review utils and add tests

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/281


src/main/webapp/app/pages/curation/review/ReviewPage.tsx
```diff
@@ -153,23 +153,25 @@ const ReviewPage = (props: IReviewPageProps) => {
...
-                handleAccept={props.acceptReviewChangeHandler}
-                handleDelete={props.rejectReviewChangeHandler}
-              />
-            ))}
-        </Col>
-      </Row>
+      {rootReview ? (
+        <Row>
+          <Col>
+            {Object.values(rootReview.children)
```
jfkonecn, line 159: The sort with the filter can be costly with memory if there's a lot of data in the object. Are you expecting to have a lot of data being processed here?

src/main/webapp/app/shared/util/firebase/firebase-review-utils.spec.ts
```diff
@@ -0,0 +1,548 @@
...
+    test.each([
+      [{ updatedBy: 'User', updateTime: 1 }, 'mutations/0/name_review', ReviewAction.NAME_CHANGE],
+      [{ updatedBy: 'User', updateTime: 1 }, 'mutations/0/tumors/0/cancerTypes_review', ReviewAction.NAME_CHANGE],
+      [{ updatedBy: 'User', updateTime: 1 }, 'mutations/0/tumors/0/TIs/0/treatments/0/name_review', ReviewAction.NAME_CHANGE],
+      [{ updatedBy: 'User', updateTime: 1, promotedToMutation: true }, 'mutations/0/name_review', ReviewAction.PROMOTE_VUS],
+      [{ updatedBy: 'User', updateTime: 1, added: true }, 'mutations/0/name_review', ReviewAction.CREATE],
+      [{ updatedBy: 'User', updateTime: 1, demotedToVus: true }, 'mutations/0/name_review', ReviewAction.DEMOTE_MUTATION],
+      [{ updatedBy: 'User', updateTime: 1, removed: true }, 'mutations/0/name_review', ReviewAction.DELETE],
+      [{ updatedBy: 'User', updateTime: 1, lastReviewed: 'old' }, 'mutations/0/description_review', ReviewAction.UPDATE],
+    ])('should return correct review action', (review, path, expected) => {
```
jfkonecn, line 53: I don't know the answer to this, but if one of these tests breaks will jest tell you which case broke? If it just gives you the test name I think you are able to use test.each to change the test name based on the data passed to it.

jfkonecn (changes requested):
Overall nice work on making the tests clear. They test one thing which is great since it will be easier to debug them if they break. I'm not sure if you did this intentionally, but you also did the best practice of structuring the test in the form of "arrange, act, assert" which also helps readability.

src/main/webapp/app/config/constants/firebase.ts
```diff
@@ -313,12 +313,16 @@ export enum ReviewLevelType {
...
+  TSG = 'Tumor Suppressor',
+  OCG = 'Oncogene',
   SUMMARY = 'Summary',
   NAME = 'Name',
   BACKGROUND = 'Background',
   PENETRANCE = 'Penetrance',
   DESCRIPTION = 'Description',
   INHERITANCE_MECHANISM = 'Mechanism of Inheritance',
+  DIAGNOSTIC = 'Diagnostic',
+  Prognostic = 'Prognostic',
```
bprize15, line 325: PROGNOSTIC

src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx
```diff
@@ -61,52 +65,47 @@ export class BaseReviewLevel {
...
-    this.lastReviewedString = lastReviewedString;
-    this.uuid = uuid;
-    this.reviewLevelType = ReviewLevelType.REVIEWABLE;
-    this.reviewAction = getReviewAction(review, reviewPath);
-    if (newState !== undefined) {
-      this.newState = newState;
-    }
-    if (oldState !== undefined) {
-      this.oldState = oldState;
+    if (!reviewInfo.reviewAction) {
```
bprize15, line 104: Wondering why need this check. I haven't seen a use case where `reviewAction` is set

src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx
```diff
@@ -147,29 +146,21 @@ export const getRelevantKeysFromUuidKey = (uuidKey: string) => {
...
 
 const removeLeafNodes = (parentReview: BaseReviewLevel) => {
   for (const key of Object.keys(parentReview.children)) {
     const childReview = parentReview.children[key];
-    if (!childReview.hasChildren()) {
-      if (childReview.reviewLevelType === ReviewLevelType.META && !(childReview as ReviewLevel).review) {
-        delete parentReview.children[key];
-      }
+    if (childReview.hasChildren()) continue;
+    if (childReview.reviewLevelType === ReviewLevelType.META) {
```
bprize15, line 162: Looks like this method is only used when you have a grouping review level (META) as a leaf. What scenario would cause this behavior?

src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx
```diff
@@ -230,41 +200,56 @@ export const reformatReviewTitle = (baseReviewLevel: BaseReviewLevel) => {
...
-export const reviewLevelSortMethod = (a: ReviewLevel, b: ReviewLevel) => {
-  const reviewActionPriority = [ReviewAction.NAME_CHANGE, ReviewAction.UPDATE, ReviewAction.CREATE, ReviewAction.DELETE];
-  const aIndex = reviewActionPriority.indexOf(a.reviewAction);
-  const bIndex = reviewActionPriority.indexOf(b.reviewAction);
-  if (aIndex === -1) {
-    return 1;
-  }
-  if (bIndex === 1) {
-    return -1;
+export const reviewLevelSortMethod = (a: BaseReviewLevel, b: BaseReviewLevel) => {
```
bprize15, line 203: Are you doing this case so that it can determine whether its a `BaseReviewLevel` or not by checking for `reviewInfo`? If so, would you consider checking the types early and doing an early return? It is slightly more verbose, but to me it is more understandable.  This is preference though, so if you disagree, feel free to disregard

bprize15 (changes requested):
One last thing. We still need to handle germline data. I remembered we talked about adding the `mutation_specific` prefix to `penetrance` to avoid confusing with gene versions.

If you want me to come back through after and make the changes for germline, I'm good with that. Or we can get it into this PR.

Screenshot 2024-04-26 at 1 08 15 PM

src/main/webapp/app/shared/util/firebase/firebase-review-utils.spec.ts
calvinlu3: It gives the test name, but in this case the names are all the same: `should return correct review action`

src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx
calvinlu3: Yeah, this can be removed

src/main/webapp/app/shared/util/firebase/firebase-review-utils.spec.ts
jfkonecn: Looks like you can string format a [name](https://jestjs.io/docs/api#1-testeachtablename-fn-timeout) for each test based on the parameters passed to the test case

jfkonecn (changes requested):

src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx
calvinlu3: When looping through an array (mutations, tumors, treatments), I create an initial META review level for each element. The reason is that, I don't know which element has additional reviewable items nested under it. 

If a mutation has no reviewable items, then it has no children, so we can remove the META review level that we created.

src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx
calvinlu3: Good point, I will try to make it explicit. 
I want to sort the `MetaReviewLevel` between Updates and Creates, but that type does not have `reviewInfo`

src/main/webapp/app/pages/curation/review/ReviewPage.tsx
jfkonecn: never mind I forgot sort returns the same array

calvinlu3:
@bprize15  I updated the germline fields with `mutation_specific` prefix.
Germline review view:
image

Another thing we need to think about is how we want to differentiate between Genomic Indicators and Mutations in review mode. Currently, we use the name of each for collapsible title, so they both look the same. We can followup with another PR.

image

bprize15 (approved):
LGTM, thanks for the changes!

PR #285, opened 2024-04-30
Support parsing simple c. and g. alteration

zhx828:
The following formats will be supported.
- c.*
- g.*
- ([0-9]{1,2}|X|Y|MT):g.*

Technically c. and g. variants will have reference allele and variant allele, but that will be supported in the future.
Screenshot 2024-05-01 at 12 43 09 AM
Screenshot 2024-05-01 at 12 43 18 AM
Screenshot 2024-05-01 at 12 54 49 AM


bprize15 (approved):
LGTM!

PR #286, opened 2024-05-01
upgrade node to 20.12.2

bprize15:


.eslintrc.json
```diff
@@ -44,6 +43,12 @@
     "@typescript-eslint/restrict-plus-operands": "off",
     "@typescript-eslint/no-floating-promises": "off",
     "@typescript-eslint/prefer-regexp-exec": "off",
+
+    // Added by Ben while upgrading node
+    "@typescript-eslint/no-unsafe-argument": "off",
+    "@typescript-eslint/no-unsafe-enum-comparison": "off",
+    "@typescript-eslint/no-misused-promises": "off",
```
zhx828, line 50: Just by looking at the name of these tracks, they all sound important. What's the rational of disabling?

zhx828 (commented):
Thank you! It looks good. Should we consider removing the package-lock json if we use yarn?

.eslintrc.json
bprize15: As discussed during standup, I checked which rules would be relatively easy to implement: 

Re-enabling `@typescript-eslint/no-unsafe-enum-comparison` was not too bad so I added that back.

`@typescript-eslint/no-unsafe-argument` disables the use of `any` types as arguments, which we do all over the place, so I left it disabled

`@typescript-eslint/no-misused-promises` checks for multiple conditions, but disallows using async functions as handlers such as many of our `onConfirm`s. Thus, I have kept the rule, but disabled errors for that condition. Here is the [documentation](https://typescript-eslint.io/rules/no-misused-promises/) if interested in checking the other conditions.

bprize15:
> Thank you! It looks good. Should we consider removing the package-lock json if we use yarn?

Created ticket to address this: https://github.com/oncokb/oncokb-pipeline/issues/346

zhx828 (approved):
LGTM, thanks!

src/main/webapp/app/components/tabs/CurationHistoryTab.tsx
```diff
@@ -123,10 +123,10 @@ const CurationHistoryTab = observer(({ historyData, usersData, addUsersListener,
 
       content.push(
         <div>
-          <Row className="justify-content-between" style={{ marginBottom: '22px' }}>
+          <div className="d-flex justify-content-between align-items-center" style={{ marginBottom: '22px' }}>
             <span>
               <TextFormat value={data.createdAt} type="date" format={APP_HISTORY_FORMAT} />
-              <span className="mr-2" />
+              <span className="me-2" />
```
calvinlu3, line 129: Looks like that starting using start and end rather than left and right to support RTL layouts?

calvinlu3 (approved):

src/main/webapp/app/components/tabs/CurationHistoryTab.tsx
bprize15: Yeah, exactly

package.json
```diff
@@ -73,129 +73,130 @@
...
     "lodash": "4.17.21",
-    "mobx": "^6.3.3",
-    "mobx-react": "^7.2.0",
+    "mobx": "^6.12.3",
+    "mobx-react": "^9.1.1",
     "oncokb-styles": "^1.5.0",
     "path-browserify": "1.0.1",
     "pluralize": "^8.0.0",
-    "rc-tooltip": "^6.1.2",
+    "rc-tooltip": "^6.2.0",
```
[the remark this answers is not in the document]
bprize15: LGTM

PR #290, opened 2024-05-02
Fix therapy dropdown resetting to default and fda level not saved properly

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/340

Issues:
1. FDA level was not saved properly in the `lastReviewed` field. It was saved as `Fda2`, but should be saved with the prefix, just `2`.
2. Noticed toggling the therapy collapsible caused the solid/liquid/fda propagation levels to reset to the default. This should only happen when highest level is changed.


zhx828 (approved):
LGTM, thanks!

PR #292, opened 2024-05-10
Gene review service tests

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/349

Ended up using https://github.com/marchaos/jest-mock-extended to help mock classes.


src/main/webapp/app/pages/curation/collapsible/ReviewCollapsible.tsx
```diff
@@ -92,7 +93,9 @@ export const ReviewCollapsible = (props: IReviewCollapsibleProps) => {
           icon={faCheck}
           color={SUCCESS}
           onClick={() => {
-            props.handleAccept(props.hugoSymbol, [props.baseReviewLevel as ReviewLevel], props.isGermline);
+            props.handleAccept(props.hugoSymbol, [props.baseReviewLevel as ReviewLevel], props.isGermline).catch(error => {
```
jfkonecn, line 96: Should the prop ever be something other than a ReviewLevel object?

src/main/webapp/app/service/firebase/firebase-gene-review-service.spec.ts
```diff
@@ -0,0 +1,340 @@
...
+          'new',
+          new Review(DEFAULT_USERNAME),
+          'uuid',
+        ),
+      ).rejects.toThrow(SentryError);
+    });
+  });
+
+  describe('acceptChanges', () => {
+    beforeEach(() => _beforeEach());
```
jfkonecn, line 98: Couldn't you put this logic at the `describe('Firebase Gene Review Service', () => {` level?

You can also be high brow about it. 🎩 

```ts
beforeEach(_beforeEach);
```

src/main/webapp/app/service/firebase/firebase-gene-service.ts
```diff
@@ -291,7 +292,7 @@ export class FirebaseGeneService {
         hugoSymbol,
         tumor.excludedCancerTypes_uuid,
         !excludedCancerTypesReview.isChangeReverted,
-        false
+        false,
```
jfkonecn, line 295: Did prettier add this? If so, why wasn't it do that before?

jfkonecn (changes requested):
Nice clean tests!

src/main/webapp/app/pages/curation/collapsible/ReviewCollapsible.tsx
calvinlu3: `props.baseReviewLevel` can be MetaReviewLevel or ReviewLevel. We still render the MetaReviewLevel in this component.

src/main/webapp/app/service/firebase/firebase-gene-service.ts
calvinlu3: Yeah, prettier added that comma after I removed it and git commited. Not sure why it wasn't added before.

src/main/webapp/app/service/firebase/firebase-gene-review-service.spec.ts
```diff
@@ -38,9 +38,9 @@ describe('Firebase Gene Review Service', () => {
     jest.useFakeTimers().setSystemTime(DEFAULT_DATE);
   };
 
-  describe('updateReviewableContent', () => {
-    beforeEach(() => _beforeEach());
+  beforeEach(_beforeEach);
```
jfkonecn, line 41: Is there a need to have the _beforeEach function as a variable if it's only being used once?

jfkonecn (approved):
I'm okay with you not addressing my comment if you disagree with it.

src/main/webapp/app/service/firebase/firebase-gene-review-service.spec.ts
calvinlu3: I agree that we can get rid of the function

src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
```diff
@@ -79,136 +58,124 @@ export class FirebaseGeneReviewService {
...
-    return this.firebaseRepository.update(getFirebaseGenePath(isGermline, hugoSymbol), updateObject).then(() => {
-      if (isChangeReverted) {
-        this.firebaseRepository.delete(`${firebasePath}_review/lastReviewed`);
-      }
-      // Update Meta information
-      this.firebaseMetaService.updateGeneMetaContent(hugoSymbol, isGermline);
-      this.firebaseMetaService.updateGeneReviewUuid(hugoSymbol, uuid, !isChangeReverted, isGermline);
-    });
+    try {
+      await this.firebaseMetaService.updateMeta(hugoSymbol, uuid, !isChangeReverted, isGermline);
```
bprize15, line 62: Assuming we didn't come up with a solution for if one of these updates failed then? I'm fine with leaving this, but if you think it's really critical to check both you can look at `createGene` in `firebase-gene-service.ts`. I created a utility method `isPromiseOk` for dealing with this sort of scenario.

src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
```diff
@@ -79,136 +58,124 @@ export class FirebaseGeneReviewService {
...
+      const { uuid, reviewAction, review, reviewPath } = reviewLevel.reviewInfo;
+
+      if (reviewAction === ReviewAction.UPDATE || reviewAction === ReviewAction.NAME_CHANGE) {
+        if (reviewLevel.nestedUnderCreateOrDelete) {
+          continue;
+        }
+        clearReview(review);
+        try {
+          await this.firebaseRepository.update(geneFirebasePath, { [reviewPath]: review });
+          await this.firebaseMetaService.updateGeneReviewUuid(hugoSymbol, uuid, false, isGermline);
```
bprize15, line 90: Same comment as above

src/main/webapp/app/oncokb-commons/components/util/NotificationUtils.tsx
```diff
@@ -36,3 +38,15 @@ export const notifyWarning = (error: OncoKBError, additionalInfo?: string, optio
 export const notifyError = (error: Error | OncoKBError, additionalInfo?: string, options?: ToastOptions) => {
   return toast.error(getErrorMessage(error, additionalInfo), options);
 };
+
+export const notifyCurationPageError = (error: Error | OncoKBError | SentryError, additionalInfo?: string) => {
```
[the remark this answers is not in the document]
calvinlu3: Merged into `notifyError`

src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
calvinlu3: I took a look at using multi-location updates (atomic operation), however it does not work well when the update depends on the data that is currently in firebase. For instance, when we accept a deletion, we need to know the current index. This use case is what `runTransaction` is used for, but in order to update both Meta and Gene collections, we need to provide the root path. `runTransaction` also reads in the data given the path, so in our case, we would be reading in our entire firebase database.

I am going to use SentryError to track errors for now.

bprize15 (approved):
LGTM, thanks for changes!

PR #299, opened 2024-05-15
Fix cancer type summary throwing exception after delete

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/361


bprize15 (approved):
LGTM, only change here (besides testing) is adding a check to make sure the tumor is not null correct? Otherwise, it seems the functionality is the same?

calvinlu3:
> LGTM, only change here (besides testing) is adding a check to make sure the tumor is not null correct? Otherwise, it seems the functionality is the same?

@bprize15 Yeah, the only change is to check if there's a tumor. I could have added `tumor?.summary`, but just wanted to return if there's no tumor.

PR #301, opened 2024-05-16
Use unique id for each radio input

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/373

fix-radio-button


zhx828 (approved):
LGTM, thanks

PR #304, opened 2024-05-16
Remove dependency on firebase User collection

zhx828:
We will use user info from the server side to manage.
This fixes https://github.com/oncokb/oncokb-pipeline/issues/354


bprize15 (approved):
LGTM

calvinlu3 (approved):

PR #306, opened 2024-05-16
Update favicon.ico

zhx828:
I don't intend to include the full favicon folder that oncokb-public use given the favicon is only used in a very minimum way.

This fixes https://github.com/oncokb/oncokb-pipeline/issues/375


calvinlu3 (approved):

calvinlu3:
@zhx828 Any idea why we had different icons before? Left is localhost, right is production:
image

zhx828:
> @zhx828 Any idea why we had different icons before? Left is localhost, right is production: image

I actually don't know.

PR #310, opened 2024-05-17
Fix newly created mutation not accepted properly

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/371

This only happens when creating a new mutation and add some data to the new mutation. There's a bug in `getAllNestedReviewUuids` method where it did not find the nested uuids, which should be cleared from Meta collection when accepting a creation.

create-save-pr


zhx828 (approved):
LGTM, thanks!

PR #316, opened 2024-05-20
fix collapsible to show correct date and action

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/383


calvinlu3 (approved):
Good catch, thanks!

PR #320, opened 2024-05-21
Sort by lastModifiedAt after needsReview in gene list page

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/386


zhx828 (approved):
Thanks

PR #327, opened 2024-05-24
Check history has records

zhx828:
This fix allows the user to go into the gene page if the history does not have record. The history without record was introduced(I think) during the review which we still need to figure out how it happened.

This is related to https://github.com/oncokb/oncokb-pipeline/issues/401


bprize15 (approved):
LGTM

PR #329, opened 2024-05-28
Always send crash error to sentry & Enable source map in production 

zhx828:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/403
@calvinlu3 you need to have SENTRY_AUTH_TOKEN in environment when building the docker image. I will slack you separately.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -164,3 +164,5 @@ Desktop.ini
 ######################
 /src/test/javascript/screenshots/actual
 /src/test/javascript/screenshots/diff
+# Sentry Config File
+.env.sentry-build-plugin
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -163,4 +163,6 @@ Desktop.ini
 # Screenshots
 ######################
 /src/test/javascript/screenshots/actual
-/src/test/javascript/screenshots/diff
\ No newline at end of file
+/src/test/javascript/screenshots/diff
+# Sentry Config File
+.env.sentry-build-plugin
```
calvinlu3, line 168: This is the env file I need to define for sentry to pickup auth token

PR #330, opened 2024-05-28
Add screenshot tests

bprize15:
The only thing to note is that the file `screenshot.ts` is a work in progress. The entire commented out section needs to be added back, but I removed it for now to isolate the SSL certificate error.


.github/workflows/webdriver-test.yml
```diff
@@ -0,0 +1,25 @@
...
+
+        - name: Build Docker images
+          run: docker-compose build
+
+        - name: Run Docker containers
+          run: |
+            docker-compose up -d
+            docker-compose run --rm wdio
+        
+        - name: Archive screenshots
```
zhx828, line 20: Do we have something similar to public https://github.com/oncokb/oncokb-public/blob/master/.github/workflows/screenshot-test.yml#L20-L31 to compare latest/base?

zhx828 (approved):
LGTM, thanks

.github/workflows/webdriver-test.yml
bprize15: We do. I save the entire screenshots folder which contains the image that was just taken, the baseline image, and the diff

PR #331, opened 2024-05-28
Fix tumor name review logic

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/405

Issues fixed:
1. Split meta uuid by `, ` to support cancer type added/updated through legacy.
2. Excluded cancer type field not cleared after accepting/rejecting

### DIFF
--- a/src/main/webapp/app/pages/curation/review/ReviewPage.tsx
+++ b/src/main/webapp/app/pages/curation/review/ReviewPage.tsx
@@ -58,7 +58,7 @@ const ReviewPage = (props: IReviewPageProps) => {
       let uuids = [];
       for (const key of Object.keys(metaReview)) {
         if (metaReview[key] === true) {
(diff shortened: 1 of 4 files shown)


src/main/webapp/app/pages/curation/review/ReviewPage.tsx
```diff
@@ -55,10 +55,10 @@ const ReviewPage = (props: IReviewPageProps) => {
 
   useEffect(() => {
     if (metaReview) {
-      const uuids = [];
+      let uuids = [];
       for (const key of Object.keys(metaReview)) {
         if (metaReview[key] === true) {
-          uuids.push(key);
+          uuids = uuids.concat(key.split(', '));
```
zhx828, line 61: I would do a split on comma and trim.

zhx828 (approved):
LGTM

PR #334, opened 2024-05-30
End keycloak session if user is not authorized

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/394

### Changes
1. If the user is unauthorized (not in our mysql database), the keycloak session should be terminated.
2. Update logged out screen to give more information

Unauthorized login:
image

Logged out:
image

### DIFF
--- a/src/main/webapp/app/pages/login/logout.tsx
+++ b/src/main/webapp/app/pages/login/logout.tsx
@@ -1,40 +1,92 @@
-import React from 'react';
-import { componentInject } from 'app/shared/util/typed-inject';
+import React, { useEffect, useMemo } from 'react';
+import { connect } from 'app/shared/util/typed-inject';
 import { IRootStore } from 'app/stores';
-import { observer } from 'mobx-react';
-import { Row } from 'reactstrap';
+import { Link, RouteComponentProps } from 'react-router-dom';
+import { PAGE_ROUTE } from 'app/config/constants/constants';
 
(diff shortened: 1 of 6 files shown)


src/main/webapp/app/pages/login/logout.tsx
```diff
@@ -1,40 +1,92 @@
...
+    const searchParams = new URLSearchParams(props.location.search);
+    return !!searchParams.get('session_terminated');
+  }, [props.location.search]);
 
-  componentDidMount() {
-    this.props.logout();
-  }
+  const unauthorizedAccess = useMemo(() => {
+    const searchParams = new URLSearchParams(props.location.search);
+    return !!searchParams.get('unauthorized');
```
zhx828, line 19: For these search params, can we leave them in a centralized place as constants?

src/main/webapp/app/pages/login/logout.tsx
```diff
@@ -1,40 +1,92 @@
...
+  }, [props.location.search]);
 
-  render() {
-    if (this.props.logoutUrl) {
-      window.location.href = this.props.logoutUrl;
+  useEffect(() => {
+    if (unauthorizedAccess && !sessionTerminated) {
```
zhx828, line 23: This logic should be in a method.

zhx828 (approved):
Looks good, minor comments.

PR #336, opened 2024-05-30
Add sort cancer types

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/372

I played around with and without debounce, and I settled on not debouncing. Since you mentioned that the API call is not an issue, I prefer the instant feedback.

### DIFF
--- a/src/main/webapp/app/components/tabs/CurationHistoryTab.tsx
+++ b/src/main/webapp/app/components/tabs/CurationHistoryTab.tsx
@@ -73,7 +73,7 @@ const CurationHistoryTab = observer(({ historyData, getUsers, users, historyTabS
 
   useEffect(() => {
     async function fetchAllDrugs() {
-      const drugs = await getDrugs({ page: 0, size: GET_ALL_DRUGS_PAGE_SIZE, sort: 'id,asc' });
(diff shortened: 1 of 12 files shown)


src/main/webapp/app/pages/curation/header/MutationsSectionHeader.tsx
```diff
@@ -342,6 +348,33 @@ function MutationsSectionHeader({
...
+                        style={{ cursor: `${isDisabled ? null : 'pointer'}`, marginLeft: '0px' }}
+                        type="checkbox"
+                      />
+                      <Label
+                        for={`tx-level-filter-${filter.label}`}
+                        style={{ cursor: `${isDisabled ? null : 'pointer'}`, marginLeft: CHECKBOX_LABEL_LEFT_MARGIN }}
+                      >
+                        {filter.label}
+                      </Label>
+                    </InputGroup>
```
zhx828, line 373: This InputGroup seems to be replicated quite a lot. Can we make it a component?

bprize15: Oh sorry about this, but I was actually working on a different PR and I accidentally forked this one off of that. This was not supposed to make it into this PR, so I just removed it, but I will definitely make it into a component for that PR

zhx828 (approved):
LGTM, thanks

PR #338, opened 2024-05-30
close mutation collapsible when not in firebase

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/408

### DIFF
--- a/src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
+++ b/src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
@@ -76,8 +76,8 @@ const MutationCollapsible = ({
 }: IMutationCollapsibleProps) => {
   const firebaseMutationsPath = `${getFirebaseGenePath(isGermline, hugoSymbol)}/mutations`;
 
-  const [mutationUuid, setMutationUuid] = useState<string>(null);
-  const [mutationName, setMutationName] = useState<string>(null);
+  const [mutationUuid, setMutationUuid] = useState<string>('');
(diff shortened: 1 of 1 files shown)


src/main/webapp/app/pages/curation/collapsible/MutationCollapsible.tsx
```diff
@@ -76,8 +76,8 @@ const MutationCollapsible = ({
 }: IMutationCollapsibleProps) => {
   const firebaseMutationsPath = `${getFirebaseGenePath(isGermline, hugoSymbol)}/mutations`;
 
-  const [mutationUuid, setMutationUuid] = useState<string>(null);
-  const [mutationName, setMutationName] = useState<string>(null);
+  const [mutationUuid, setMutationUuid] = useState<string>('');
+  const [mutationName, setMutationName] = useState<string>('');
```
calvinlu3, line 80: Why change to empty string?

bprize15: I don't want to render anything if the data is not available yet. This can happen for two reasons: 

1. The mutation isn't present (it was deleted) 
2. It just has not finished loading yet

We previously only handled the first case. I did this by checking `!mutationUuid || !mutationName`. However, I need to make the distinction between not done loading and it doesn't exist. 

If it is not done loading, my previous condition will still work.

However, if it does not exist both values will be set to `null`. In this case I can use `_.isNil` to check for this.

calvinlu3 (approved):

PR #341, opened 2024-06-04
add more screenshot tests

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/367


src/test/javascript/screenshots/baseline/desktop_chrome/vus-table--1920x1080.png
calvinlu3, line 1: I would try to change the mock data to include a combination of red and yellow badges as well as non-outdated VUS.

src/test/javascript/screenshots/baseline/desktop_chrome/mutation-collapsible-info--1920x1080.png
calvinlu3, line 1: Looks cut off at the top

bprize15: Changed to mutation collapsible based on our conversation

webpack/webpack.common.js
```diff
@@ -108,6 +110,8 @@ module.exports = async options => {
             SERVER_API_URL: `''`,
             DOCKER: options.DOCKER,
           },
+          'Date.now': process.env.DOCKER ? `() => ${DEFAULT_DATE}` : 'Date.now',
+          'new Date()': process.env.DOCKER ? `new Date(${DEFAULT_DATE})` : 'new Date()',
```
calvinlu3, line 114: Will this be picked up if we run tests locally and not through docker?

bprize15: This will not. I think it's just an edge case that we should leave alone (and not introduce additional complexity). The tests will fail anyway if run locally since the screen size will be off. The only reason I would ever run the tests locally is during development so you don't need to keep rebuilding the image. Running the tests locally should always be expected to fail.

src/main/webapp/app/pages/curation/collapsible/BaseCollapsible.tsx
```diff
@@ -37,6 +37,13 @@ export interface BaseCollapsibleProps {
   showLoadingSpinner?: boolean;
 }
 
+function getDataTestid(dataTestid: string, identifier: React.ReactNode) {
```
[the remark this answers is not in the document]
bprize15: We allow a React.ReactNode to be passed in as the collapsible title for greater flexibility. In this case, I just do a check to make sure it's a string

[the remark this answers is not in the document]
calvinlu3: Thanks for clarifying. I wonder what's the best practice to assigning `data-testid`. 

Is it better to manually generate unique test ids for each element in a list-like component.
Or, should we query for all element with `data-testid='sample-id'` (which will return a list) and use indices to get the one you're looking for? (ie in react-testing-library, there is a function `getAllByTestId`)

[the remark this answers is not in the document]
bprize15: I don't think we need the extra step of querying all elements with a testid and then using the index. Seems straightforward enough to just use the title. Could even use the uuid if we want, but I think that might be overkill, especially since we control the test data.

calvinlu3 (approved):

PR #314, opened 2024-05-20
Fix germline review

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/381
resolves https://github.com/oncokb/oncokb-pipeline/issues/379

PR #343, opened 2024-06-05
Meta review levels should not have actions/info

calvinlu3:
Issue:
image

This issue was introduced with the change in the screenshot: (https://github.com/oncokb/oncokb-transcript/pull/314)
image

We should add a test case to https://github.com/oncokb/oncokb-pipeline/issues/366


bprize15 (approved):
LGTM

PR #345, opened 2024-06-06
change textareas to autoresize

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/421


src/main/webapp/app/shared/firebase/input/RealtimeBasicInput.tsx
```diff
@@ -139,6 +139,12 @@ const RealtimeBasicInput: React.FunctionComponent<IRealtimeBasicInput> = (props:
   const inputComponent = (
     <>
       <Input
+        innerRef={inputRef => {
+          if (inputRef && type === RealtimeInputType.TEXTAREA) {
+            inputRef.style.height = 'auto';
+            inputRef.style.height = `${inputRef.scrollHeight}px`;
```
zhx828, line 145: We overwrite the height twice?

bprize15: The first overwrite is for the case when the text box needs to shrink. When a row is deleted, the scroll height will not automatically change. That set to `auto` causes it to shrink to fit it's content.

There second one is for when the box grows. In that case the scroll height will automatically increase, and so I just need to set the height to the scroll height.

zhx828: Interesting, so basically after assigning auto, the inputref height is updated right away so you can get the height in the second assignment?

zhx828 (approved):
LGTM, the webdriver test is failing though.

calvinlu3 (approved):
Thanks for changes!

Just want to mention that in the `/diff` folder, there are two screenshots that I don't really see a diff.

PR #346, opened 2024-06-07
change cancer type query to prioritize starts with

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/372


zhx828 (approved):
Works great! Thanks!

PR #349, opened 2024-06-09
Create a reference parsing method instead of regex

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/359

### DIFF
--- a/src/main/webapp/app/config/constants/constants.ts
+++ b/src/main/webapp/app/config/constants/constants.ts
@@ -424,3 +424,5 @@ export const DEFAULT_TOAST_ERROR_OPTIONS: ToastOptions = {
 export const KEYCLOAK_LOGOUT_REDIRECT_PARAM = 'post_logout_redirect_uri';
 export const KEYCLOAK_SESSION_TERMINATED_PARAM = 'session_terminated';
 export const KEYCLOAK_UNAUTHORIZED_PARAM = 'unauthorized';
+
+export const REFERENCE_IDENTIFIERS = ['PMID: ', 'Abstract: ', 'NCT'];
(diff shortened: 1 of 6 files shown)


src/main/webapp/app/config/constants/constants.ts
```diff
@@ -424,3 +424,5 @@ export const DEFAULT_TOAST_ERROR_OPTIONS: ToastOptions = {
 export const KEYCLOAK_LOGOUT_REDIRECT_PARAM = 'post_logout_redirect_uri';
 export const KEYCLOAK_SESSION_TERMINATED_PARAM = 'session_terminated';
 export const KEYCLOAK_UNAUTHORIZED_PARAM = 'unauthorized';
+
+export const REFERENCE_IDENTIFIERS = ['PMID: ', 'Abstract: ', 'NCT'];
```
zhx828, line 428: We used to can support `PMID` `PMID:` `PMID ` because of the flexibility of the regex. Now we have to have `PMID: ` in the text. I don't think this is ok.

zhx828:
@calvinlu3 JS does not have recursive regex. But we can support one more level with parenthesis with `\(\s*(?:PMID|NCT|Abstract):?.*?(?:\([^()]*\).*?)*\)` how many levels we need to support?

calvinlu3:
@zhx828 At the moment, we've only encountered 1 nested parenthesis. We can go with that updated regex until we see a use case that doesn't work with it.

src/main/webapp/app/config/constants/constants.ts
calvinlu3: That's good to know. Always thought it had one type of format since most are written in that way. It can be fixed with a regex.

calvinlu3:
Closing this PR in favor of https://github.com/oncokb/oncokb-transcript/pull/356 because we are still planning to use REGEX

calvinlu3:
Closing this PR in favor of https://github.com/oncokb/oncokb-transcript/pull/356 because we are still planning to use REGEX

PR #350, opened 2024-06-10
Use google diff-match-patch to compare the text new/old content

zhx828:
## Tabs diff
Screenshot 2024-06-10 at 2 03 31 PM
Screenshot 2024-06-10 at 2 28 09 PM

## Stack diff 
Screenshot 2024-06-10 at 4 44 59 PM

## Merged diff
Screenshot 2024-06-10 at 4 24 15 PM

This fixes https://github.com/oncokb/oncokb-pipeline/issues/420


calvinlu3 (commented):
@zhx828
- The textareas will fit the full text once the feature branch is rebased with the latest rc (Ben's changes). I think that's the behavior we want.
- I haven't gotten a chance to checkout locally. Looking at the code, it doesn't seem like the `new` content is ever updated to Firebase if a user is editing the text area.

src/main/webapp/app/components/diff-viewer/DiffViewer.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  return (
+    <Tabs
+      className={'mx-0'}
+      tabs={[
+        {
+          title: 'New',
+          content: (
+            <div>
+              {typeof props.new === 'string' && <Input type={'textarea'} value={props.new} style={{ backgroundColor: '#f1f9ff80' }} />}
```
bprize15, line 59: In the case where `props.new === 'string`, I assume the intention is just to display the data and not allow editing. Maybe we could set 'disabled' for the input field or just us a span instead?

zhx828:
> I haven't gotten a chance to checkout locally. Looking at the code, it doesn't seem like the new content is ever updated to Firebase if a user is editing the text area.
@calvinlu3 I'm using the RealtimeTextArea. It's supposed to update the content.

src/main/webapp/app/components/diff-viewer/DiffViewer.tsx
zhx828: Agree. I will play around with span vs disabled.

zhx828: I decided to do a disabled textarea instead of showing span. Using span makes the content section and diff section to similar

 
Screenshot 2024-06-12 at 8 29 56 AM
Screenshot 2024-06-12 at 8 30 23 AM

zhx828:
@calvinlu3 the tests are failing currently, I see the same errors in the base branch. I will level them alone. Let me know once you update your branch, I will make sure they are passing before merging.

bprize15 (approved):
LGTM, thanks!

calvinlu3 (approved):
@zhx828 Tests in the base branch is passing now. You can rebase.
I am skipping the Webdriver screenshots, since they will be updated before this feature branch is merged.

zhx828:
Move the pr to https://github.com/oncokb/oncokb-transcript/pull/361

PR #351, opened 2024-06-11
Add variant_recommendation table

zhx828:


src/main/resources/config/liquibase/changelog/20240611000000_added_entity_VariantRecommendation.xml
```diff
@@ -0,0 +1,35 @@
...
+    <changeSet id="20240611000000-1" author="HZ">
+        <createTable tableName="variant_recommendation">
+            <column name="id" type="bigint" autoIncrement="true">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="entrez_gene_id" type="integer">
+                <constraints nullable="true" />
+            </column>
+            <column name="hugo_symbol" type="varchar(255)">
+                <constraints nullable="true" />
```
jfkonecn, line 18: These are the latest column I've seen in the merged version of the statistics file. 
`Hugo_Symbol` `HGVSp_Short` `criteria_all` `criteria_caner_type` `criteria_detailed_cancer_type`

`entrez_gene_id` and `variant` are missing from the last iteration of the statistics file I've seen.
`HGVSp_Short` is missing in this table definition.

jfkonecn (changes requested):

src/main/resources/config/liquibase/changelog/20240611000000_added_entity_VariantRecommendation.xml
zhx828: @jfkonecn HGVSp_Short is variant in this case. We should include entrez_gene_id in the stats file if available.

jfkonecn (approved):

PR #355, opened 2024-06-12
Move action buttons to collapsibles with content to review

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/414

### DIFF
--- a/src/main/webapp/app/pages/curation/review/ReviewPage.tsx
+++ b/src/main/webapp/app/pages/curation/review/ReviewPage.tsx
@@ -46,6 +46,8 @@ const ReviewPage: React.FunctionComponent<IReviewPageProps> = (props: IReviewPag
 
   const [splitView, setSplitView] = useState(false);
 
+  /* eslint-disable no-console */
+
   useEffect(() => {
     if (geneEntity && props.firebaseInitSuccess) {
       // Fetch the data when the user enters review mode. We don't use a listener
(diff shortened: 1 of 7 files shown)


src/main/webapp/app/pages/curation/review/ReviewPage.tsx
```diff
@@ -40,20 +42,23 @@ const ReviewPage: React.FunctionComponent<IReviewPageProps> = (props: IReviewPag
...
   const [splitView, setSplitView] = useState(false);
 
+  const fetchFirebaseData = () => {
+    // Fetch the data when the user enters review mode. We don't use a listener
+    // because there shouldn't be another user editing the gene when it is being reviewed.
+    get(ref(props.firebaseDb, firebaseGenePath)).then(snapshot => setGeneData(snapshot.val()));
+    get(ref(props.firebaseDb, firebaseMetaReviewPath)).then(snapshot => setMetaReview(snapshot.val()));
+  };
+
   useEffect(() => {
```
bprize15, line 58: Probably a slight nit, but you could move the `props.GetDrugs` call to a separate `useEffect`. That way, you could save an API call or two if firebase is not yet ready

src/main/webapp/app/pages/curation/review/ReviewPage.tsx
```diff
@@ -86,28 +90,45 @@ const ReviewPage: React.FunctionComponent<IReviewPageProps> = (props: IReviewPag
...
   };
 
-  const markRootCollapsibleAsPending = (reviewLevelId: string) => {
-    const newReviewChildren = reviewChildren.map(c => {
-      if (c.id === reviewLevelId) {
-        c.hideLevel = true;
-      }
-      return c;
-    });
+  const deleteCollapsible = (reviewLevelId: string, isPending = false) => {
```
bprize15, line 110: Maybe either separate into two methods or add a comment so it's clear that this method is called twice

bprize15 (commented):
This could be personal preference, but I wonder if we could add the `Created` flag next to the nested cancer type. For me, it seems to improve readability, but I could imagine this not being an issue for curators, so wondering your thoughts?

Screenshot 2024-06-14 at 4 30 27 PM

src/main/webapp/app/pages/curation/review/ReviewPage.tsx
calvinlu3: Separated logic into `deleteCollapsible` and `hideCollapsible`.

bprize15 (approved):
LGTM, thanks!

PR #356, opened 2024-06-13
Update reference Regex to support 1 level of nested parenthesis

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/359


calvinlu3:
@zhx828 You said that we also need to update the Regex in oncokb-core, but I am assuming you mean oncokb-public because we don't parse the text for references in the core (at least I could not identify the code).

If the code looks good to you here, I will open another PR to update the regex for public website and I think we also need to update for cbioportal frontend

zhx828 (approved):
LGTM

PR #358, opened 2024-06-14
Replace npm with yarn

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/346


calvinlu3 (commented):
Changes look good to me. Seems like Github Hosted Runners will have yarn installed by default. 

@bprize15 
- Do you plan on removing the package-lock?
- The `.npmrc` file is a configuration file used by npm. Looks like we can remove that?

bprize15:
> Changes look good to me. Seems like Github Hosted Runners will have yarn installed by default.
> 
> @bprize15
> 
> * Do you plan on removing the package-lock?
> * The `.npmrc` file is a configuration file used by npm. Looks like we can remove that?

Deleted `package-lock` but looks like yarn classic (1.x) works with `.npmrc` (https://verdaccio.org/docs/setup-yarn/)

calvinlu3 (approved):

zhx828 (approved):
LGTM. Thank you.

PR #360, opened 2024-06-17
Add button to toggle between somatic and germline in gene list page

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/445

### Changes
- Store germline/somatic setting in localStorage
- Add toggle button to gene list page
   - I removed Validation and History download tabs when in Germline setting we're not releasing Germline genes yet.
- Redirect to `/somatic` by default, unless localStorage has germline setting stored.


src/main/webapp/app/pages/curation/GeneListPage.tsx
```diff
@@ -80,51 +95,61 @@ const GeneListPage = (props: StoreProps) => {
...
+
+  if (!isGermline) {
+    sidebarTabs.push({
+      title: 'Data Validation',
+      content: <CurationDataValidationTab />,
+    });
+    sidebarTabs.push({
+      title: 'History',
+      content: <ReviewHistoryTab />,
+    });
```
zhx828, line 113: We are not showing history?

src/main/webapp/app/config/constants/constants.ts
```diff
@@ -427,3 +432,5 @@ export const DEFAULT_TOAST_ERROR_OPTIONS: ToastOptions = {
 export const KEYCLOAK_LOGOUT_REDIRECT_PARAM = 'post_logout_redirect_uri';
 export const KEYCLOAK_SESSION_TERMINATED_PARAM = 'session_terminated';
 export const KEYCLOAK_UNAUTHORIZED_PARAM = 'unauthorized';
+
+export const SOMATIC_GERMLINE_SETTING_KEY = 'oncokbCurationSomaticGermlineSettingKey';
```
zhx828, line 436: We need an util to manage how local storage keys are generated.

zhx828 (changes requested):

src/main/webapp/app/pages/curation/GeneListPage.tsx
calvinlu3: The history tab is only used to generate the README of changes for the data release.

zhx828: Yeah, but I would still prefer to see the history if it's possible.

src/main/webapp/app/config/constants/constants.ts
```diff
@@ -427,3 +432,10 @@ export const DEFAULT_TOAST_ERROR_OPTIONS: ToastOptions = {
 export const KEYCLOAK_LOGOUT_REDIRECT_PARAM = 'post_logout_redirect_uri';
 export const KEYCLOAK_SESSION_TERMINATED_PARAM = 'session_terminated';
 export const KEYCLOAK_UNAUTHORIZED_PARAM = 'unauthorized';
+
+/**
+ * LocalStorage Keys
+ * Use getLocalStorageKey util method to attach application name to key
+ */
+export const PRIORITY_ENTITY_MENU_ITEM_KEY = 'entityMenuPriorityKey';
+export const SOMATIC_GERMLINE_SETTING_KEY = 'somaticGermlineSettingKey';
```
calvinlu3, line 441: @zhx828 I think the naming for localStorage keys should be `${appName}-{key}` to reduce the possibility of name clashes across different applications.

Also, do you prefer a util method or defining all localstorage key in constants with a comment with how they should be named? I find it easy to accidently not use the `getLocalStorageKey` util.

zhx828: @calvinlu3 agree with the suggestion on the key format.
Personally I think we should have a local storage store which it will manage all the keys. Fine for me to use constants for now.

src/main/webapp/app/pages/curation/GeneListPage.tsx
```diff
@@ -24,39 +28,50 @@ type GeneMetaInfo = {
...
+  const pathname = props.location.pathname;
+  const isGermline = pathname.includes(GERMLINE_PATH);
+
   useEffect(() => {
     if (props.firebaseReady) {
-      const unsubscribe = props.addMetaListener(FB_COLLECTION.META);
-      return () => unsubscribe && unsubscribe();
+      const callbacks = [];
+      if (isGermline) {
+        callbacks.push(props.addGermlineMetaListener(FB_COLLECTION.GERMLINE_META));
```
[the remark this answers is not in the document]
calvinlu3: The firebase data store only stores a single `data` corresponding to the path. To re-use, we probably have to support an array of data?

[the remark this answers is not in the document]
bprize15: True, but do we ever need access to both germline data and somatic data at the same time? I assume when the page is toggled, the correct path will be used and the store will be overwritten with the correct data.

PR #365, opened 2024-06-19
Feature/strict

jfkonecn:
closes oncokb/oncokb-pipeline#332

### DIFF
--- a/src/main/webapp/app/app.tsx
+++ b/src/main/webapp/app/app.tsx
@@ -14,18 +14,19 @@ import AppRoutes from 'app/routes/routes';
 import NavigationSidebar from 'app/components/sidebar/NavigationSidebar';
 import Layout from './layout';
 import LoadingIndicator, { LoaderSize } from 'app/oncokb-commons/components/loadingIndicator/LoadingIndicator';
+import { Unsubscribe } from 'firebase/database';
 
-const baseHref = document.querySelector('base').getAttribute('href').replace(/\/$/, '');
+const baseHref = document.querySelector('base')?.getAttribute('href')?.replace(/\/$/, '');
 
 export type IAppProps = StoreProps;
 
 const App: React.FunctionComponent<IAppProps> = (props: IAppProps) => {
   useEffect(() => {
-    let authSubscriber = undefined;
+    let authSubscriber: Unsubscribe | undefined = undefined;
     if (props.isCurator) {
       authSubscriber = props.initializeFirebase();
     }
-    return () => authSubscriber && authSubscriber();
+    return () => authSubscriber?.();
   }, [props.isCurator]);
 
   return (
@@ -58,7 +59,7 @@ const App: React.FunctionComponent<IAppProps> = (props: IAppProps) => {
 const mapStoreToProps = ({ authStore, layoutStore, firebaseAppStore }: IRootStore) => ({
   isAuthorized: authStore.isAuthorized,
   authorities: authStore.account.authorities,
-  isCurator: hasAnyAuthority(authStore.account.authorities, [AUTHORITIES.CURATOR]),
+  isCurator: hasAnyAuthority(authStore.account.authorities ?? [], [AUTHORITIES.CURATOR]),
   loadingAuth: authStore.loading,
   navigationSidebarWidth: layoutStore.navigationSidebarWidth,
   toggleNavSidebar: layoutStore.toggleNavigationSidebar,
--- a/src/main/webapp/app/components/geneHistoryTooltip/gene-history-tooltip-utils.tsx
+++ b/src/main/webapp/app/components/geneHistoryTooltip/gene-history-tooltip-utils.tsx
@@ -6,7 +6,9 @@ import { IDrug } from 'app/shared/model/drug.model';
 import { FlattenedHistory } from 'app/shared/util/firebase/firebase-history-utils';
 import DiffViewer from 'app/components/diff-viewer/DiffViewer';
 
-export default function constructTimeSeriesData(record: FlattenedHistory): RequiredTimeSeriesEventData | ExtraTimeSeriesEventData {
+export default function constructTimeSeriesData(
+  record: FlattenedHistory,
+): RequiredTimeSeriesEventData | ExtraTimeSeriesEventData | undefined {
   let operation: string;
   let bubbleColor: string;
   let content: React.ReactNode;
@@ -62,7 +64,7 @@ export default function constructTimeSeriesData(record: FlattenedHistory): Requi
   };
 }
 
-export function getTimeSeriesDataContent(newContent: string, oldContent?: string) {
+export function getTimeSeriesDataContent(newContent: string | undefined, oldContent?: string) {
   return (
     <div className="gene-history-event-content">
       <DiffViewer new={newContent} old={oldContent} type={'merged'} />
--- a/src/main/webapp/app/components/panels/CompanionDiagnosticDevicePanel.tsx
+++ b/src/main/webapp/app/components/panels/CompanionDiagnosticDevicePanel.tsx
@@ -147,4 +147,4 @@ const mapStoreToProps = ({ associationStore }: IRootStore) => ({});
 
 type StoreProps = ReturnType<typeof mapStoreToProps>;
 
-export default connect<object, StoreProps>(mapStoreToProps)(CompanionDiagnosticDevicePanel);
+export default connect(mapStoreToProps)(CompanionDiagnosticDevicePanel);
--- a/src/main/webapp/app/components/panels/CompanionDiagnosticDevicePanel.tsx
+++ b/src/main/webapp/app/components/panels/CompanionDiagnosticDevicePanel.tsx
@@ -6,18 +6,19 @@ import { useHistory, useLocation } from 'react-router-dom';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faPlus, width } from '@fortawesome/free-solid-svg-icons/faPlus';
 import DefaultTooltip from 'app/shared/tooltip/DefaultTooltip';
-import CancerTypeSelect from 'app/shared/select/CancerTypeSelect';
+import CancerTypeSelect, { CancerTypeSelectOption } from 'app/shared/select/CancerTypeSelect';
 import { SaveButton } from 'app/shared/button/SaveButton';
 import GeneSelect from 'app/shared/select/GeneSelect';
 import AlterationSelect from 'app/shared/select/AlterationSelect';
-import DrugSelect from 'app/shared/select/DrugSelect';
+import DrugSelect, { DrugSelectOption } from 'app/shared/select/DrugSelect';
 import FdaSubmissionSelect from 'app/shared/select/FdaSubmissionSelect';
 import { connect } from 'app/shared/util/typed-inject';
 import { IRootStore } from 'app/stores';
 import { getEntityActionRoute } from 'app/shared/util/RouteUtils';
-import { Association, CancerType } from 'app/shared/api/generated/curation';
+import { Alteration, Association, CancerType, Drug, FdaSubmission } from 'app/shared/api/generated/curation';
 import { associationClient } from 'app/shared/api/clients';
 import { notifyError, notifySuccess } from 'app/oncokb-commons/components/util/NotificationUtils';
+import { IFdaSubmission } from 'app/shared/model/fda-submission.model';
 
 const SidebarMenuItem: React.FunctionComponent<{ style?: React.CSSProperties; children: React.ReactNode }> = ({ style, children }) => {
   return <div style={{ padding: '0.5rem 1rem', ...style }}>{children}</div>;
@@ -29,11 +30,12 @@ export const defaultAdditional = {
 };
 
 const CompanionDiagnosticDevicePanel: React.FunctionComponent<StoreProps> = props => {
-  const [selectedGeneId, setSelectedGeneId] = useState(null);
-  const [alterationValue, onAlterationChange] = useState(null);
-  const [cancerTypeValue, onCancerTypeChange] = useState(null);
-  const [drugValue, onDrugChange] = useState(null);
-  const [fdaSubmissionValue, onFdaSubmissionChange] = useState(null);
+  const [selectedGeneId, setSelectedGeneId] = useState<string | null>(null);
+  type AlterationChangeArg = Parameters<NonNullable<Parameters<typeof AlterationSelect>[0]['onChange']>>[0];
+  const [alterationValue, onAlterationChange] = useState<AlterationChangeArg | null>(null);
+  const [cancerTypeValue, onCancerTypeChange] = useState<CancerTypeSelectOption | null>(null);
+  const [drugValue, onDrugChange] = useState<readonly DrugSelectOption[]>();
+  const [fdaSubmissionValue, onFdaSubmissionChange] = useState<readonly IFdaSubmission[]>();
 
   const history = useHistory();
   const location = useLocation();
@@ -42,25 +44,36 @@ const CompanionDiagnosticDevicePanel: React.FunctionComponent<StoreProps> = prop
   const createBiomarkerAssociation = (e: any) => {
     e.preventDefault();
     const association: Association = {
-      fdaSubmissions: fdaSubmissionValue.map(fdaSubmission => {
+      fdaSubmissions: fdaSubmissionValue?.map((fdaSubmission): FdaSubmission => {
         return {
-          id: fdaSubmission.value,
+          // TYPE-ISSUE: value is not on FdaSubmission
+          id: (fdaSubmission as any).value,
+          number: '',
+          curated: false,
+          genetic: false,
+          deviceName: '',
+          supplementNumber: '',
         };
       }),
-      alterations: alterationValue.map(alteration => {
+      alterations: alterationValue?.map((alteration): Alteration => {
         return {
           id: alteration.value,
+          type: 'UNKNOWN',
+          name: '',
+          alteration: '',
+          proteinChange: '',
         };
       }),
-      drugs: drugValue.map(drug => {
+      drugs: drugValue?.map((drug): Drug => {
         return {
           id: drug.value,
+          uuid: '',
         };
       }),
     };
 
     if (cancerTypeValue) {
-      association.cancerTypes = [{ id: cancerTypeValue.value as number } as CancerType];
+      association.cancerTypes = [{ id: cancerTypeValue.value } as CancerType];
     }
     associationClient
       .createAssociation(association)
@@ -86,16 +99,17 @@ const CompanionDiagnosticDevicePanel: React.FunctionComponent<StoreProps> = prop
           <SidebarMenuItem>Add Biomarker Association</SidebarMenuItem>
           <SidebarMenuItem>
             <GeneSelect
+              isMulti={false}
               onChange={option => {
                 const geneId = option ? option.value : null;
-                setSelectedGeneId(geneId);
+                setSelectedGeneId(geneId?.toString() ?? null);
               }}
             />
           </SidebarMenuItem>
           <SidebarMenuItem>
             <div style={{ display: 'flex' }}>
               <div style={{ flex: 1 }}>
-                <AlterationSelect isMulti geneId={selectedGeneId} onChange={onAlterationChange} />
+                <AlterationSelect isMulti geneId={selectedGeneId ?? ''} onChange={onAlterationChange} />
               </div>
               <DefaultTooltip overlay={'Create new alteration'}>
                 <Button className="ms-1" color="primary" onClick={redirectToCreateAlteration}>
@@ -105,7 +119,7 @@ const CompanionDiagnosticDevicePanel: React.FunctionComponent<StoreProps> = prop
             </div>
           </SidebarMenuItem>
           <SidebarMenuItem>
-            <CancerTypeSelect onChange={onCancerTypeChange} />
+            <CancerTypeSelect isMulti={false} onChange={onCancerTypeChange} />
           </SidebarMenuItem>
           <SidebarMenuItem>
             <DrugSelect isMulti onChange={onDrugChange} />
@@ -135,4 +149,4 @@ const mapStoreToProps = ({ associationStore }: IRootStore) => ({});
 
 type StoreProps = ReturnType<typeof mapStoreToProps>;
 
-export default connect(mapStoreToProps)(CompanionDiagnosticDevicePanel);
+export default connect<object, StoreProps>(mapStoreToProps)(CompanionDiagnosticDevicePanel);
--- a/src/main/webapp/app/components/panels/CompanionDiagnosticDevicePanel.tsx
+++ b/src/main/webapp/app/components/panels/CompanionDiagnosticDevicePanel.tsx
@@ -97,7 +97,6 @@ const CompanionDiagnosticDevicePanel: React.FunctionComponent<StoreProps> = prop
           <SidebarMenuItem>Add Biomarker Association</SidebarMenuItem>
           <SidebarMenuItem>
             <GeneSelect
-              isMulti={false}
               onChange={option => {
                 const geneId = option ? option.value : null;
                 setSelectedGeneId(geneId?.toString() ?? null);
@@ -117,7 +116,7 @@ const CompanionDiagnosticDevicePanel: React.FunctionComponent<StoreProps> = prop
             </div>
           </SidebarMenuItem>
           <SidebarMenuItem>
-            <CancerTypeSelect isMulti={false} onChange={onCancerTypeChange} />
+            <CancerTypeSelect onChange={onCancerTypeChange} />
           </SidebarMenuItem>
           <SidebarMenuItem>
             <DrugSelect isMulti onChange={onDrugChange} />
--- a/src/main/webapp/app/components/panels/CompanionDiagnosticDevicePanel.tsx
+++ b/src/main/webapp/app/components/panels/CompanionDiagnosticDevicePanel.tsx
@@ -9,9 +9,9 @@ import DefaultTooltip from 'app/shared/tooltip/DefaultTooltip';
 import CancerTypeSelect, { CancerTypeSelectOption } from 'app/shared/select/CancerTypeSelect';
 import { SaveButton } from 'app/shared/button/SaveButton';
 import GeneSelect from 'app/shared/select/GeneSelect';
-import AlterationSelect from 'app/shared/select/AlterationSelect';
+import AlterationSelect, { AlterationSelectOption } from 'app/shared/select/AlterationSelect';
 import DrugSelect, { DrugSelectOption } from 'app/shared/select/DrugSelect';
-import FdaSubmissionSelect from 'app/shared/select/FdaSubmissionSelect';
+import FdaSubmissionSelect, { FdaSubmissionSelectOption } from 'app/shared/select/FdaSubmissionSelect';
 import { connect } from 'app/shared/util/typed-inject';
 import { IRootStore } from 'app/stores';
 import { getEntityActionRoute } from 'app/shared/util/RouteUtils';
@@ -31,11 +31,10 @@ export const defaultAdditional = {
 
 const CompanionDiagnosticDevicePanel: React.FunctionComponent<StoreProps> = props => {
   const [selectedGeneId, setSelectedGeneId] = useState<string | null>(null);
-  type AlterationChangeArg = Parameters<NonNullable<Parameters<typeof AlterationSelect>[0]['onChange']>>[0];
-  const [alterationValue, onAlterationChange] = useState<AlterationChangeArg | null>(null);
+  const [alterationValue, onAlterationChange] = useState<readonly AlterationSelectOption[] | null>(null);
   const [cancerTypeValue, onCancerTypeChange] = useState<CancerTypeSelectOption | null>(null);
   const [drugValue, onDrugChange] = useState<readonly DrugSelectOption[]>();
-  const [fdaSubmissionValue, onFdaSubmissionChange] = useState<readonly IFdaSubmission[]>();
+  const [fdaSubmissionValue, onFdaSubmissionChange] = useState<readonly FdaSubmissionSelectOption[]>();
 
   const history = useHistory();
   const location = useLocation();
@@ -46,8 +45,7 @@ const CompanionDiagnosticDevicePanel: React.FunctionComponent<StoreProps> = prop
     const association: Association = {
       fdaSubmissions: fdaSubmissionValue?.map((fdaSubmission): FdaSubmission => {
         return {
-          // TYPE-ISSUE: value is not on FdaSubmission
-          id: (fdaSubmission as any).value,
+          id: fdaSubmission.value,
           number: '',
           curated: false,
           genetic: false,
--- a/src/main/webapp/app/components/sidebar/NavigationSidebar.tsx
+++ b/src/main/webapp/app/components/sidebar/NavigationSidebar.tsx
@@ -169,7 +169,7 @@ const MenuDivider: React.FunctionComponent = () => {
   return <div className="nav-menu-divider"></div>;
 };
 
-export const NavigationSidebar: React.FunctionComponent<StoreProps> = props => {
+export const NavigationSidebar: React.FunctionComponent<StoreProps> = ({ isNavSidebarCollapsed = false, ...props }) => {
   const [entityMenuFrequencies, setEntityMenuFrequencies] = useState(getDefaultEntityMenuFrequencies() as EntityMenuFrequency[]);
   const [entityMenuOrder, setEntityMenuOrder] = useState(DEFAULT_ENTITY_MENU_ORDER);
 
@@ -211,11 +211,11 @@ export const NavigationSidebar: React.FunctionComponent<StoreProps> = props => {
   }, []);
 
   return (
-    <Sidebar collapsed={props.isNavSidebarCollapsed} width={`${SIDEBAR_EXPANDED_WIDTH}px`} collapsedWidth={`${SIDEBAR_COLLAPSED_WIDTH}px`}>
+    <Sidebar collapsed={isNavSidebarCollapsed} width={`${SIDEBAR_EXPANDED_WIDTH}px`} collapsedWidth={`${SIDEBAR_COLLAPSED_WIDTH}px`}>
       <div style={{ display: 'flex', flexDirection: 'column', height: '100%' }}>
         <div style={{ lineHeight: '5rem', display: 'flex', justifyContent: 'center' }}>
           <NavbarBrand tag={NavLink} to={PAGE_ROUTE.HOME}>
-            {props.isNavSidebarCollapsed ? (
+            {isNavSidebarCollapsed ? (
               <OptimizedImage height={30} src={oncokbSmallLogo} alt={'OncoKB'} />
             ) : (
               <OptimizedImage height={30} src={oncokbLogo} alt={'OncoKB'} />
@@ -233,7 +233,7 @@ export const NavigationSidebar: React.FunctionComponent<StoreProps> = props => {
           <Menu>
             {props.isCurator && (
               <MenuItemCollapsible
-                isCollapsed={props.isNavSidebarCollapsed ?? false}
+                isCollapsed={isNavSidebarCollapsed}
                 text="Curation"
                 icon={<FiFileText size={DEFAULT_NAV_ICON_SIZE} />}
                 nav={<NavLink to={PAGE_ROUTE.CURATION} />}
@@ -242,7 +242,7 @@ export const NavigationSidebar: React.FunctionComponent<StoreProps> = props => {
             {props.isUser && (
               <>
                 <MenuItemCollapsible
-                  isCollapsed={props.isNavSidebarCollapsed ?? false}
+                  isCollapsed={isNavSidebarCollapsed}
                   text={'Search'}
                   icon={<BiSearchAlt size={DEFAULT_NAV_ICON_SIZE} />}
                   nav={<NavLink to={PAGE_ROUTE.SEARCH} />}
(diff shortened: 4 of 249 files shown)


src/main/webapp/app/components/panels/CompanionDiagnosticDevicePanel.tsx
```diff
@@ -42,25 +44,36 @@ const CompanionDiagnosticDevicePanel: React.FunctionComponent<StoreProps> = prop
   const createBiomarkerAssociation = (e: any) => {
     e.preventDefault();
     const association: Association = {
-      fdaSubmissions: fdaSubmissionValue.map(fdaSubmission => {
+      fdaSubmissions: fdaSubmissionValue?.map((fdaSubmission): FdaSubmission => {
         return {
-          id: fdaSubmission.value,
+          // TYPE-ISSUE: value is not on FdaSubmission
+          id: (fdaSubmission as any).value,
```
calvinlu3, line 50: I think in `FdaSubmissionSelect` component, we need a type for `FdaSubmissionSelectOptions` with `value` and `label`.

The type currently is set to `IFdaSubmission`, which is not the correct type

src/main/webapp/app/components/sidebar/OncoKBSidebar.tsx
```diff
@@ -80,18 +80,18 @@ const OncoKBSidebarExpanded = observer(
...
 
         if (newWidth <= ONCOKB_SIDEBAR_MIN_WIDTH) {
-          setOncoKBSidebarWidth(ONCOKB_SIDEBAR_MIN_WIDTH);
+          setOncoKBSidebarWidth?.(ONCOKB_SIDEBAR_MIN_WIDTH);
         } else if (newWidth >= maxWidth) {
-          setOncoKBSidebarWidth(maxWidth);
+          setOncoKBSidebarWidth?.(maxWidth);
         } else {
-          setOncoKBSidebarWidth(oncoKBSidebarWidth + startX - mouseMoveEvent.pageX);
+          setOncoKBSidebarWidth?.((oncoKBSidebarWidth ?? 0) + startX - mouseMoveEvent?.pageX);
```
calvinlu3, line 94: How come `mouseMoveEvent` needs optional chaining, but not on line 83? If it can really be undefined, then the function definition should probably reflect that?

src/main/webapp/app/components/tabs/CancerRiskTabs.tsx
```diff
@@ -15,9 +15,12 @@ export interface ICancerRiskTabsProps extends StoreProps {
 }
 
 function CancerRiskTabs({ cancerRiskPath, firebaseDb, textAreaClass }: ICancerRiskTabsProps) {
-  const [cancerRisk, setCancerRisk] = useState<CancerRisk>(null);
+  const [cancerRisk, setCancerRisk] = useState<CancerRisk>();
```
calvinlu3, line 18: Not sure what preference is, but I usually explicitly state the initial value for useState. (without any arguments, by default its undefined).

src/main/webapp/app/components/tabs/CurationToolsTab.tsx
```diff
@@ -12,6 +12,7 @@ import { CURRENT_REVIEWER } from 'app/config/constants/constants';
 import { GeneType } from 'app/shared/model/firebase/firebase.model';
 import { onValue, ref } from 'firebase/database';
 import { FB_COLLECTION } from 'app/config/constants/firebase';
+import { Unsubscribe } from 'firebase/auth';
```
calvinlu3, line 15: I think the type should come from `firebase/database` instead

src/main/webapp/app/entities/cancer-type/cancer-type-update.tsx
```diff
@@ -45,12 +41,13 @@ export const CancerTypeUpdate = (props: ICancerTypeUpdateProps) => {
     }
   }, [updateSuccess]);
 
+  // TYPE-ISSUE: I don't know what type values is
```
calvinlu3, line 44: It technically should be the type of the entity (ie `ICancerType`), but the values come from the `ValidatedField`, which can be any type?

calvinlu3 (changes requested):
I've only look through about half the files. Will continue tomorrow, but here are my comments so far.
1. I think all `StoreProps` can be redefined to `type StoreProps = ReturnType<typeof mapStoreToProps>` because those are always injected and available to the component. Not sure why we had `Partial`
2. I remember @zhx828 mentioned this, but our frontend models (files with `.model.ts`) make all our properties optional. However, in the backend we do have NonEmpty or NonNullable constraints. If we can match with the backend, then we can skip a lot of the optional chaining for properties that will never be undefined.

src/main/webapp/app/entities/cancer-type/cancer-type-update.tsx
calvinlu3: Same goes for all the `saveEntity` methods. The type should be the entity model type.

zhx828:
> I've only look through about half the files. Will continue tomorrow, but here are my comments so far.
> 
> 1. I think all `StoreProps` can be redefined to `type StoreProps = ReturnType<typeof mapStoreToProps>` because those are always injected and available to the component. Not sure why we had `Partial`
> 2. I remember @zhx828 mentioned this, but our frontend models (files with `.model.ts`) make all our properties optional. However, in the backend we do have NonEmpty or NonNullable constraints. If we can match with the backend, then we can skip a lot of the optional chaining for properties that will never be undefined.

Since we decided to drop jhipster frontend support. We might as well update these models by ourselves to correlate with the backend models. @jfkonecn @calvinlu3 I think it's better to do that in this pr?

calvinlu3 (changes requested):

src/main/webapp/app/components/geneHistoryTooltip/gene-history-tooltip-utils.tsx
```diff
@@ -6,7 +6,9 @@ import { IDrug } from 'app/shared/model/drug.model';
 import { FlattenedHistory } from 'app/shared/util/firebase/firebase-history-utils';
 import DiffViewer from 'app/components/diff-viewer/DiffViewer';
 
-export default function constructTimeSeriesData(record: FlattenedHistory): RequiredTimeSeriesEventData | ExtraTimeSeriesEventData {
+export default function constructTimeSeriesData(
+  record: FlattenedHistory,
+): RequiredTimeSeriesEventData | ExtraTimeSeriesEventData | undefined {
```
zhx828, line 11: Now I wonder, should this method return undefined? @bprize15 when the record operation is unknown, it seems to me we should still return a event data.

calvinlu3:
I think a lot of the changes requested have to do with improper types. Maybe, whoever more familiar with certain parts of the app can give proper types in a separate PR?

src/main/webapp/app/components/panels/CompanionDiagnosticDevicePanel.tsx
```diff
@@ -86,16 +99,17 @@ const CompanionDiagnosticDevicePanel: React.FunctionComponent<StoreProps> = prop
           <SidebarMenuItem>Add Biomarker Association</SidebarMenuItem>
           <SidebarMenuItem>
             <GeneSelect
+              isMulti={false}
```
zhx828, line 102: This is new to me. Why are we making the prop required when it's not in react-select?

src/main/webapp/app/components/sidebar/NavigationSidebar.tsx
```diff
@@ -232,7 +233,7 @@ export const NavigationSidebar: React.FunctionComponent<StoreProps> = props => {
           <Menu>
             {props.isCurator && (
               <MenuItemCollapsible
-                isCollapsed={props.isNavSidebarCollapsed}
+                isCollapsed={props.isNavSidebarCollapsed ?? false}
```
zhx828, line 236: For optional props within functional component, we should define a default instead of doing multiple checks in the same component.

src/main/webapp/app/components/sidebar/NavigationSidebar.tsx
```diff
@@ -289,9 +290,9 @@ const mapStoreToProps = ({ layoutStore, authStore }: IRootStore) => ({
   isNavSidebarCollapsed: layoutStore.isNavigationSidebarCollapsed,
   navigationSidebarWidth: layoutStore.navigationSidebarWidth,
   isAuthenticated: authStore.isAuthenticated,
-  isAdmin: hasAnyAuthority(authStore.account.authorities, [AUTHORITIES.ADMIN]),
-  isCurator: hasAnyAuthority(authStore.account.authorities, [AUTHORITIES.CURATOR]),
-  isUser: hasAnyAuthority(authStore.account.authorities, [AUTHORITIES.USER]),
+  isAdmin: hasAnyAuthority(authStore.account.authorities ?? [], [AUTHORITIES.ADMIN]),
+  isCurator: hasAnyAuthority(authStore.account.authorities ?? [], [AUTHORITIES.CURATOR]),
+  isUser: hasAnyAuthority(authStore.account.authorities ?? [], [AUTHORITIES.USER]),
   account: authStore.account,
```
zhx828, line 296: See my other reply comment to Calvin. Once we fix the backend models. These checks won't be necessary.

src/main/webapp/app/components/sidebar/OncoKBSidebar.tsx
```diff
@@ -51,7 +51,7 @@ const OncoKBSidebarExpanded = observer(
 
     const sidebarRef = useRef<HTMLDivElement>(null);
     const draggableRef = useRef<HTMLDivElement>(null);
-    const sidebarHighlightTimer = useRef(null);
+    const sidebarHighlightTimer = useRef<NodeJS.Timeout>();
```
zhx828, line 54: We are changing the type here?

zhx828 (commented):
These are my comments after going through half of the PR. Could you also run prettier one more time before merging? Just want to make sure all files are formatted, some of the changes in this PR should've been captured by prettier when commit. Not sure why they are not.
I will continue reviewing.

src/main/webapp/app/entities/drug/drug-update.tsx
```diff
@@ -133,12 +131,15 @@ export const DrugUpdate = (props: IDrugUpdateProps) => {
...
-                />
+                {drugEntity.nciThesaurus && (
+                  <NcitCodeSelect
+                    isMulti={false}
+                    ncit={drugEntity.nciThesaurus}
+                    onChange={selectedOption => {
+                      setSelectedNcit(selectedOption ? parseNcitUniqId(selectedOption.value) : undefined);
+                    }}
+                  />
+                )}
```
zhx828, line 142: This changes the intended functionality. The page is also served when adding a new entity which we need to allow user to select ncit code.

src/main/webapp/app/entities/drug/drug.tsx
```diff
@@ -41,12 +41,14 @@ export const Drug = (props: IDrugProps) => {
   const getUniqueGenes = (associations: IAssociation[]) => {
     const biomarkers: DrugAssocBiomarkers = {};
     associations?.forEach(val => {
-      for (const alteration of val.alterations) {
+      for (const alteration of val?.alterations ?? []) {
         const geneName = getGeneNameFromAlteration(alteration);
```
zhx828, line 45: getGeneNameFromAlteration should not return undefined when alteration is available. If alteration genes list is empty, the currently method returns empty string. We should keep it that way.

src/main/webapp/app/stores/firebase/firebase-data.store.ts
```diff
@@ -16,12 +16,20 @@ export class FirebaseDataStore<T> {
...
   addListener = (path: string) => {
-    const callback = (snapshot: DataSnapshot) => (this.data = snapshot.val());
-    const unsubscribe = onValue(ref(this.firebaseAppStore.firebaseDb, path), action(callback));
-    return unsubscribe;
+    if (this.firebaseAppStore.firebaseDb) {
+      const callback = (snapshot: DataSnapshot) => (this.data = snapshot.val());
+      const unsubscribe = onValue(ref(this.firebaseAppStore.firebaseDb, path), action(callback));
+      return unsubscribe;
+    } else {
+      throw new Error('No firebaseDb');
```
bprize15, line 24: We throw this error a lot. Maybe we can make it its own class

src/main/webapp/app/shared/util/typed-inject.ts
```diff
@@ -2,12 +2,28 @@ import { inject, useObserver } from 'mobx-react';
 import { IRootStore } from 'app/stores';
 import { MobXProviderContext } from 'mobx-react';
 import React from 'react';
+import { Props as SelectProps } from 'react-select';
 
-export const connect = (grabStoresFn: (param: IRootStore | Record<string, never>, any) => any) => {
-  return baseComponent => {
-    const component = ownProps => {
+/**
```
bprize15, line 7: Assuming we just want to remove this?

src/main/webapp/app/shared/util/jhipster-types.ts
```diff
@@ -1,7 +1,7 @@
 import { AxiosResponse } from 'axios';
 import { CancellablePromise } from 'mobx/dist/api/flow';
 
-export type IPayload<T> = Promise<AxiosResponse<T>> | CancellablePromise<unknown>;
+export type IPayload<T> = Promise<AxiosResponse<T>> | CancellablePromise<AxiosResponse<T>>;
```
bprize15, line 4: Why did jhipster use `unknown` instead of `AxiosResponse<T>` before?

jfkonecn: No idea, but seems like type T is what should be expected based on the code I saw

src/main/webapp/app/app.tsx
```diff
@@ -58,7 +59,7 @@ const App: React.FunctionComponent<IAppProps> = (props: IAppProps) => {
 const mapStoreToProps = ({ authStore, layoutStore, firebaseAppStore }: IRootStore) => ({
   isAuthorized: authStore.isAuthorized,
   authorities: authStore.account.authorities,
-  isCurator: hasAnyAuthority(authStore.account.authorities, [AUTHORITIES.CURATOR]),
+  isCurator: hasAnyAuthority(authStore.account.authorities ?? [], [AUTHORITIES.CURATOR]),
```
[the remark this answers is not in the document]
jfkonecn: I use `??` since it's a check for `null` or `undefined` instead of doing a "falsy" because that's normally what you want anyway. So for example you could have a check on a number like this `0 || 1` which will result in the value being 1. With `0 ?? 1` the result is 0.

src/main/webapp/app/components/panels/CompanionDiagnosticDevicePanel.tsx
jfkonecn: It's to help the type checker know that we are using single values instead of arrays. You could get the same effect by defaulting a value for isMulti in the GeneSelect component itself.

src/main/webapp/app/components/sidebar/NavigationSidebar.tsx
jfkonecn: Yeah that makes sense. My mind wasn't thinking so clearly after hours of doing this. 😄

src/main/webapp/app/shared/select/CancerTypeSelect.tsx
```diff
@@ -24,7 +24,7 @@ export type CancerTypeSelectOption = {
 };
 
 const getAllMainTypes = (cancerTypeList: ICancerType[]) => {
-  return _.uniq(cancerTypeList.filter(cancerType => cancerType.level <= 0)).sort();
+  return _.uniq(cancerTypeList.filter(cancerType => cancerType.level && cancerType.level <= 0)).sort();
```
bprize15, line 27: Here is an example of something that cannot be null (according to our database rules). As Calvin mentioned, a better solution would be to fix our models.

src/main/webapp/app/shared/modal/AddMutationModal.tsx
```diff
@@ -151,24 +157,24 @@ function AddMutationModal({
...
+        proteinEnd: alteration?.proteinEnd === -1 ? undefined : alteration?.proteinEnd,
         refResidues: alteration?.refResidues,
         varResidues: alteration?.varResidues,
       };
     }
 
     async function setExistingAlterations() {
-      if (mutationToEdit.alterations?.length > 0) {
-        setTabStates(mutationToEdit.alterations.map(alt => convertAlterationToAlterationData(alt)));
+      if ((mutationToEdit?.alterations?.length ?? 0) > 0) {
```
bprize15, line 168: Nit: Not necessary since `null > 0 === false`

src/main/webapp/app/shared/util/typed-inject.ts
jfkonecn: Good catch. I'll double check that.

src/main/webapp/app/service/firebase/firebase-meta-service.ts
```diff
@@ -73,7 +73,7 @@ export class FirebaseMetaService {
     return Promise.resolve();
   };
 
-  updateGeneCurrentReviewer = async (hugoSymbol: string, isGermline: boolean, isReviewing: boolean) => {
+  updateGeneCurrentReviewer = async (hugoSymbol: string, isGermline: boolean | undefined, isReviewing: boolean) => {
```
bprize15, line 76: I don't expect `isGermline` to be undefined

src/main/webapp/app/components/geneHistoryTooltip/gene-history-tooltip-utils.tsx
bprize15: Based on my implementation from a while back, I handle an unknown operation by returning undefined and then parsing it out. This bit could probably use a refactor, but I think it makes sense given the current implementation.

src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx
```diff
@@ -489,13 +506,13 @@ export const buildNameReview = (
 
   const valuePathParts = [currValuePath, nameKey];
   const valuePath = valuePathParts.join('/');
-  const title = readableName;
+  const title = readableName ?? '';
```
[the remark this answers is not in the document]
bprize15: Looks like this PR changed the definition of `name` for `Mutation` to `name: string | undefined = ''`. Given that we default name to an empty string, I think we should revert this change.

src/main/webapp/app/components/timeSeries/TimeSeries.tsx
```diff
@@ -96,7 +96,7 @@ const TimeSeriesInfo = ({ date, children }: ITimeSeriesInfoProps) => {
 
   useEffect(() => {
     function updateLineHeight() {
-      setLineHeight(timelineRef.current.clientHeight - lastChildRef.current.clientHeight - 10); // 10 is height of margin from react-event-timeline
+      setLineHeight((timelineRef.current?.clientHeight ?? 0) - (lastChildRef.current?.clientHeight ?? 0) - 10); // 10 is height of margin from react-event-timeline
```
bprize15, line 99: Can we set the height to null when the ref is undefined. Setting it to -10 seems like it would cause a bug

src/main/webapp/app/components/tabs/history-tab.store.tsx
```diff
@@ -26,7 +26,7 @@ export class HistoryTabStore {
     });
   }
 
-  setSelectedAuthor(author: { label: string; value: string }) {
+  setSelectedAuthor(author: { label: string; value: string } | null) {
```
bprize15, line 29: While `selectedAuthor` can be null, I don't want this method to be able to set it to null

src/main/webapp/app/components/tabs/ReviewHistoryTab.tsx
```diff
@@ -28,14 +31,14 @@ function ReviewHistoryTab({ isGermline, firebaseDb, drugList, getDrugs }: IRevie
...
 
   useEffect(() => {
-    getDrugs({ page: 0, size: GET_ALL_DRUGS_PAGE_SIZE, sort: ['id,asc'] });
+    getDrugs?.({ page: 0, size: GET_ALL_DRUGS_PAGE_SIZE, sort: ['id,asc'] });
   }, []);
 
   function handleDownload() {
-    const start = startDate ? new Date(startDate) : null;
-    const end = endDate ? new Date(endDate) : null;
+    const start = startDate ? new Date(startDate) : undefined;
```
bprize15, line 38: This is an example of a misuse of the `Partial` type. In this case, `StoreProps` should be changed to `ReturnType<typeof mapStoreToProps>`. Then the type system won't complain about things such as `getDrugs` that should never be null

src/main/webapp/app/components/tabs/GeneListPageToolsTab.tsx
```diff
@@ -35,7 +35,9 @@ function GeneListPageToolsTab({ metaData, createGene }: IGeneListPageToolsTab) {
   }
 
   async function handleCreateGene() {
-    await createGene(selectedGene.current, false, `${PAGE_ROUTE.CURATION}/${selectedGene.current}/somatic`);
+    if (selectedGene.current !== undefined) {
```
bprize15, line 38: Misuse of `Partial` keyword

src/main/webapp/app/components/tabs/CurationToolsTab.tsx
```diff
@@ -137,10 +144,10 @@ export function CurationToolsTab({
       } else {
         newGene.flags = [newFlag];
       }
-      await updateGene(newGene);
-      await searchGenes({ query: geneName, exact: true }); // repopulate gene store entities
+      await updateGene?.(newGene);
```
bprize15, line 147: Misuse of `Partial`

bprize15 (changes requested):
It looks like we misused the `Partial` type a bunch with our `StoreProps`. A bunch of these checks could be avoided if we remove that type.

src/main/webapp/app/shared/util/typed-inject.ts
jfkonecn: The `git diff` being weird. I added types to this function, but it's counting it as a delete followed by an add.

src/main/webapp/app/shared/modal/AddMutationModal.tsx
jfkonecn: That's not a nit. It's a fair comment.
The strict checking counts greater than/less than logic as an error if a `null`/`undefined` is involved.

src/main/webapp/app/service/firebase/firebase-meta-service.ts
jfkonecn: Let's revisit this comment after I fix the database types

src/main/webapp/app/components/timeSeries/TimeSeries.tsx
jfkonecn: can do

src/main/webapp/app/components/tabs/history-tab.store.tsx
jfkonecn: I'll fix it

src/main/webapp/app/components/tabs/GeneListPageToolsTab.tsx
jfkonecn: What do you mean by that?

src/main/webapp/app/app.tsx
```diff
@@ -14,18 +14,19 @@ import AppRoutes from 'app/routes/routes';
...
 
-const baseHref = document.querySelector('base').getAttribute('href').replace(/\/$/, '');
+const baseHref = document.querySelector('base')?.getAttribute('href')?.replace(/\/$/, '');
 
 export type IAppProps = StoreProps;
 
 const App: React.FunctionComponent<IAppProps> = (props: IAppProps) => {
   useEffect(() => {
-    let authSubscriber = undefined;
+    let authSubscriber: Unsubscribe | undefined = undefined;
```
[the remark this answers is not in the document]
jfkonecn: I don't think so, but if you could prove me wrong that would simplify this PR.
I agree do though, theoretically the type could be inferred.

jfkonecn:
> These are my comments after going through half of the PR. Could you also run prettier one more time before merging? Just want to make sure all files are formatted, some of the changes in this PR should've been captured by prettier when commit. Not sure why they are not. I will continue reviewing.



I just ran it. These are the files that need formatting. It doesn't look like the file I touch aren't formatted correctly.
```sh
src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
src/main/java/org/mskcc/oncokb/curation/aop/logging/LoggingAspect.java
src/main/java/org/mskcc/oncokb/curation/config/OAuth2Configuration.java
src/main/java/org/mskcc/oncokb/curation/importer/CdxImporter.java
src/main/java/org/mskcc/oncokb/curation/importer/GeneImporter.java
src/main/java/org/mskcc/oncokb/curation/importer/Importer.java
src/main/java/org/mskcc/oncokb/curation/importer/MetaImporter.java
src/main/java/org/mskcc/oncokb/curation/importer/OncoTreeImporter.java
src/main/java/org/mskcc/oncokb/curation/importer/TranscriptImporter.java
src/main/java/org/mskcc/oncokb/curation/security/jwt/TokenProvider.java
src/main/java/org/mskcc/oncokb/curation/service/AlterationQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/ArticleQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/CancerTypeService.java
src/main/java/org/mskcc/oncokb/curation/service/ClinicalTrialArmQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/ClinicalTrialQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/CompanionDiagnosticDeviceQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/ConsequenceQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/DrugQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/EligibilityCriteriaQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/EnsemblGeneQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/EvidenceQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/FdaDrugQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/FdaDrugService.java
src/main/java/org/mskcc/oncokb/curation/service/FdaSubmissionQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/FlagQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/GeneQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/GenomeFragmentQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/GenomeNexusService.java
src/main/java/org/mskcc/oncokb/curation/service/GenomicIndicatorQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/MainService.java
src/main/java/org/mskcc/oncokb/curation/service/NciThesaurusQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/NihEutilsService.java
src/main/java/org/mskcc/oncokb/curation/service/SequenceQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/SynonymQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/TranscriptQueryService.java
src/main/java/org/mskcc/oncokb/curation/service/TranscriptService.java
src/main/java/org/mskcc/oncokb/curation/service/UserService.java
src/main/java/org/mskcc/oncokb/curation/service/mapper/TranscriptMapper.java
src/main/java/org/mskcc/oncokb/curation/service/mapper/UserMapper.java
src/main/java/org/mskcc/oncokb/curation/util/AlterationUtils.java
src/main/java/org/mskcc/oncokb/curation/util/CancerTypeUtils.java
src/main/java/org/mskcc/oncokb/curation/util/CdxUtils.java
src/main/java/org/mskcc/oncokb/curation/util/FileUtils.java
src/main/java/org/mskcc/oncokb/curation/util/TimeUtil.java
src/main/java/org/mskcc/oncokb/curation/web/filter/OAuth2RefreshTokensWebFilter.java
src/main/java/org/mskcc/oncokb/curation/web/rest/AlleleStateResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/AlterationResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/ArticleResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/AssociationResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/CancerTypeResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/CategoricalAlterationResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/ClinicalTrialArmResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/ClinicalTrialResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/CompanionDiagnosticDeviceResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/ConsequenceResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/DrugResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/EligibilityCriteriaResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/EnsemblGeneResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/EvidenceResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/FdaDrugResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/FdaSubmissionResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/FdaSubmissionTypeResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/FlagResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/GeneResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/GenomeFragmentResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/GenomicIndicatorResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/InfoResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/LevelOfEvidenceResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/NciThesaurusResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/RuleResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/SeqRegionResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/SequenceResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/SpecimenTypeResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/SynonymResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/TranscriptController.java
src/main/java/org/mskcc/oncokb/curation/web/rest/TranscriptResource.java
src/main/java/org/mskcc/oncokb/curation/web/rest/errors/ExceptionTranslator.java
src/main/resources/templates/error.html
src/main/webapp/404.html
src/main/webapp/app/config/icon-loader.ts
src/main/webapp/app/index.tsx
src/main/webapp/app/shared/api/generated/core/api.ts
src/main/webapp/app/shared/api/generated/core/base.ts
src/main/webapp/app/shared/api/generated/core/common.ts
src/main/webapp/swagger-ui/index.html
src/test/java/org/mskcc/oncokb/curation/config/TestSecurityConfiguration.java
src/test/java/org/mskcc/oncokb/curation/service/MailServiceIT.java
src/test/java/org/mskcc/oncokb/curation/web/rest/AccountResourceIT.java
src/test/java/org/mskcc/oncokb/curation/web/rest/ClientForwardControllerTest.java
src/test/java/org/mskcc/oncokb/curation/web/rest/LogoutResourceIT.java
```

jfkonecn:
> I've only look through about half the files. Will continue tomorrow, but here are my comments so far.
> 
> 1. I think all `StoreProps` can be redefined to `type StoreProps = ReturnType<typeof mapStoreToProps>` because those are always injected and available to the component. Not sure why we had `Partial`
> 2. I remember @zhx828 mentioned this, but our frontend models (files with `.model.ts`) make all our properties optional. However, in the backend we do have NonEmpty or NonNullable constraints. If we can match with the backend, then we can skip a lot of the optional chaining for properties that will never be undefined.

@zhx828  and @calvinlu3 for number 1 if I remove the Partial from the type definition then all the injected props become required. I tried to get the types to work, but I couldn't figure it out. I'm going to move on to other fixes for this feature brach and try again after I fix everything else.

src/main/webapp/app/components/sidebar/OncoKBSidebar.tsx
jfkonecn: Good catch! It didn't need that.

src/main/webapp/app/components/tabs/CancerRiskTabs.tsx
jfkonecn: I generally go with `undefined` to represent no value since that's the javascript generally represents an empty value.

I prefer `useState<CancerRisk>()` over `useState<CancerRisk | null>(null)` since they basically get you the same result, but `useState<CancerRisk>()` you have to type less and there's less to read. Also since null is a special kind of empty value in JavaScript sometimes it is meaningful when you set something to null over undefined which can be confusing in some situations.

src/main/webapp/app/components/tabs/CurationToolsTab.tsx
jfkonecn: Makes sense. I make that mistake in too many places. 😄

src/main/webapp/app/components/panels/CompanionDiagnosticDevicePanel.tsx
zhx828: @jfkonecn the react-select does not make this required field, we also should not. Could you elaborate why adding this would help with type checker?

jfkonecn: @zhx828 basically removing `isMulti` makes typescript confused about what the underlying value. It thinks it can be either a single value or an array if you leave `isMulti` undefined which is clearly wrong. I'll see if I can make it work without the property.

src/main/webapp/app/entities/cancer-type/cancer-type-update.tsx
jfkonecn: for line 50 there is a call to `parentId` like so `values.parentId.toString()` which doesn't exist on the `ICancerType` interface.

src/main/webapp/app/entities/genomic-indicator/genomic-indicator.tsx
```diff
@@ -24,8 +24,8 @@ const getAssociatedAlterations = (associations?: IAssociation[]) => {
   );
 };
 
-const getAssociatedAlleleStates = (alleleStates?: IAlleleState[]) => {
-  return alleleStates?.map(alleleState => alleleState.name).join();
+const getAssociatedAlleleStates = (alleleStates: IAlleleState[] | undefined) => {
```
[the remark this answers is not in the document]
jfkonecn: We absolutely can use an optional parameter, but if you do it this way it forces the caller to include the argument. If you think it's unnecessary then I'd be happy to revert it.

src/main/webapp/app/pages/curation/review/ReviewPage.tsx
```diff
@@ -180,7 +184,7 @@ const ReviewPage: React.FunctionComponent<IReviewPageProps> = (props: IReviewPag
         <Row>
           <Col>
             <ReviewCollapsible
-              hugoSymbol={hugoSymbol}
+              hugoSymbol={hugoSymbol ?? ''}
```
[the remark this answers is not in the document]
jfkonecn: good catch. I'll fixed it the model changes PR.

src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
```diff
@@ -100,11 +106,16 @@ export class FirebaseGeneReviewService {
...
           const tumorReviewLevel = reviewLevel as TumorReviewLevel;
-          const excludedCtReviewPath = tumorReviewLevel.excludedCancerTypesReviewInfo.reviewPath;
-          updateObject[excludedCtReviewPath] = review;
+          const excludedCtReviewPath = tumorReviewLevel.excludedCancerTypesReviewInfo?.reviewPath;
+          if (excludedCtReviewPath) {
+            updateObject[excludedCtReviewPath] = review;
+          }
+        }
+        if (!uuid) {
+          throw new SentryError('UUID is missing', { hugoSymbol, reviewLevel, isGermline });
```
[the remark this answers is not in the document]
jfkonecn: There are some parts of the code review, uuid, and lastReviewString could be `undefined` base on the type information. So I changed `ReviewInfo` to reflect that.
[here's an example](https://github.com/oncokb/oncokb-transcript/blob/57042ed46d9c92c8e3c0d3212857285f5c4d5eba/src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx#L624-L627)

src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx
```diff
@@ -210,37 +215,47 @@ export const removeLeafNodes = (parentReview: BaseReviewLevel) => {
...
+      return false;
+    }
   }
   return false;
 };
 
 export const getCompactReviewInfo = (review: BaseReviewLevel) => {
   if (review.reviewLevelType === ReviewLevelType.REVIEWABLE_MULTI) return review;
-  if (!review.hasChildren()) {
+  if (!review.hasChildren?.()) {
```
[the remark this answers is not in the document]
jfkonecn: good catch!

src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx
```diff
@@ -688,7 +705,7 @@ export const buildStringReview = (
 };
 
 export const buildObjectReview = (
-  obj: Record<string, any>,
+  obj: object | Implication | null,
```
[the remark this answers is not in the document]
jfkonecn: I got it down to just `obj: object`. I don't remember why I added `Implication` but I removed it too. As far as null goes, there's a check [here](https://github.com/oncokb/oncokb-transcript/blob/rc/src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx#L436) that looks to see if the type of the object is `object`. Since we're in the world of JavaScript null is an a typeof object so it doesn't guard against null. I added a null check to that if statement too.

jfkonecn:
Creating a new feature branch

PR #367, opened 2024-06-19
Introduce Data Import tab

zhx828:
### Add a new tab in genes page
Screenshot 2024-06-24 at 11 31 30 PM

### Allow to choose gene type (we probably should use our somatic/germline toggle. I can switch after review page feature is merged)
Screenshot 2024-06-19 at 5 18 11 PM

### Allow to choose data type and group by different level(I only added gene level data import for demo, will all more)
Screenshot 2024-06-19 at 5 18 24 PM

### Indicate import status and provide message
Screenshot 2024-06-24 at 11 31 43 PM

@calvinlu3 @bprize15 want to get some feedbacks before adding tests
This is related to https://github.com/oncokb/oncokb-pipeline/issues/353


src/main/webapp/app/components/tabs/CurationDataImportTab.tsx
```diff
@@ -0,0 +1,449 @@
...
+const getOptionalColumns = (dataType: DataImportType) => {
+  switch (dataType) {
+    case DataImportType.GENE_GENOMIC_INDICATOR:
+      return ['entrez_gene_id', 'description', 'alleleState'];
+    default:
+      return [];
+  }
+};
+
+const geneCheck = async (firebaseGeneService: FirebaseGeneService, isGermline: boolean, hugoSymbol: string, callback) => {
```
bprize15, line 85: Small nit, but could we change `callback` to something more descriptive, maybe `validateData`? Could we also give it the type `() => Promise<DataImportStatus>`?

src/main/webapp/app/components/tabs/CurationDataImportTab.tsx
```diff
@@ -0,0 +1,449 @@
...
+      return ['hugo_symbol', 'value'];
+    default:
+      return [];
+  }
+};
+
+const getOptionalColumns = (dataType: DataImportType) => {
+  switch (dataType) {
+    case DataImportType.GENE_GENOMIC_INDICATOR:
+      return ['entrez_gene_id', 'description', 'alleleState'];
```
bprize15, line 79: I don't see `entrez_gene_id` or `alleleState` being used elsewhere. Do we currently plan on supporting importing these fields?

bprize15 (commented):
Since the tabs now need to be scrolled, the sidebar is broken and cannot be dragged when it is at its minimum size. However, I played around and you can fix this by replacing `width: 10px` with `min-width: 10px` and `max-width: 10px` in `oncokb-sidebar.scss` in the `draggable` class.

src/main/webapp/app/components/tabs/CurationDataImportTab.tsx
```diff
@@ -0,0 +1,449 @@
...
+      return [];
+  }
+};
+
+const geneCheck = async (firebaseGeneService: FirebaseGeneService, isGermline: boolean, hugoSymbol: string, callback) => {
+  const status: DataImportStatus = {
+    status: 'initialized',
+    message: '',
+  };
+  hugoSymbol = hugoSymbol.trim();
```
calvinlu3, line 90: I think FB paths are case sensitive, so the import might not work if the hugoSymbol is not cased exactly the same as firebase.

src/main/webapp/app/components/tabs/CurationDataImportTab.tsx
```diff
@@ -0,0 +1,449 @@
...
+  fileReader.onload = event => {
+    setFileUploadErrorMsg('');
+    const text = event.target.result as string;
+    tsvFileToArray(text);
+  };
+
+  const requiredCols: string[] = selectedDataType !== undefined ? getRequiredColumns(selectedDataType) : [];
+  const optionalCols: string[] = selectedDataType !== undefined ? getOptionalColumns(selectedDataType) : [];
+
+  // reset import statues after changing gene type and data type
```
calvinlu3, line 230:
```suggestion
  // reset import status after changing gene type and data type
```

calvinlu3 (changes requested):
1. Can we show the filename that was chosen?
2. After clicking `Import`, should we disable the button? I think there's no point to click it again before the user re-uploads the file.

src/main/webapp/app/components/tabs/CurationDataImportTab.tsx
zhx828: I will get rid of entrez_gene_id. My intention was to avoid outdated hugo symbol. But currently I'm not planning to make an API to server side to verify the hugo symbol. The hugo symbol has to be the latest one.

zhx828: `allele_state` I will support

src/main/webapp/app/components/tabs/CurationDataImportTab.tsx
bprize15: Yeah, that makes sense to me too

src/main/webapp/app/components/tabs/CurationDataImportTab.tsx
zhx828: Yeah, let's keep it this way.

src/main/webapp/app/service/firebase/firebase-gene-service.ts
```diff
@@ -410,17 +411,57 @@ export class FirebaseGeneService {
...
 
     newGenomicIndicator.associationVariants = [{ name: PATHOGENIC_VARIANTS, uuid: PATHOGENIC_VARIANTS }];
 
     const newReview = new Review(this.authStore.fullName);
     newReview.updateTime = new Date().getTime();
-    newReview.added = true;
-    newGenomicIndicator.name_review = newReview;
 
-    await this.firebaseRepository.pushToArray(genomicIndicatorsPath, [newGenomicIndicator]);
+    newGenomicIndicator.name = name;
```
bprize15, line 434: It looks like you do a bunch of work to create the review even if `skipReview === true`. The resources required for this work are definitely negligible, but can we only do it if `skipReview === false`?

src/test/javascript/spec/e2e.ts
```diff
@@ -78,3 +100,118 @@ describe('Screenshot Tests', () => {
...
+
+  const uploadGenomicIndicatorToImport = async () => {
+    await selectGenomicIndicatorToImport();
+
+    const fileToUpload = path.join(process.cwd(), 'src/test/javascript/data/genomic_indicators.tsv');
+    const remoteFilePath = await browser.uploadFile(fileToUpload);
+    await $(`input[id=${DATA_IMPORT_FILE_UPLOAD_INPUT_ID}]`).setValue(remoteFilePath);
+  };
+
+  it('File required and optional columns alerts are shown', async () => {
```
bprize15, line 172: We should move this to screenshot tests

src/test/javascript/spec/e2e.ts
```diff
@@ -78,3 +100,118 @@ describe('Screenshot Tests', () => {
...
+    await expandIcon.click();
+
+    // too complicated to check logics, do a screenshot instead
+    const dataTable = await $(`div[id=${DATA_IMPORT_DATA_TABLE_ID}]`);
+    expect(dataTable).toExist();
+    const result = await browser.checkElement(dataTable, 'data-import-data-table-before-import', methodOptions);
+    expect(result).toBeLessThanOrEqual(0);
+  });
+
+  it('Data properly imported and show warning/error if available', async () => {
```
bprize15, line 196: I am a little more hesitant to add this one to screenshot tests since it checks if the import button is disabled. However, if it needs to save a screenshot I am leaning toward putting it in screenshot tests.

bprize15 (changes requested):
I noticed that I am able to select `Germline`, select `Genomic Indicator`, and then switch back to `Somatic`. Maybe we should make it so the dropdown is automatically deselected in this case?

src/test/javascript/spec/e2e.ts
zhx828: Agree

src/main/webapp/app/service/firebase/firebase-gene-service.ts
zhx828: I changed the name to toReivew so I don't have to do a lot of `!skipReview`

bprize15 (approved):
LGTM, thanks!

src/main/webapp/app/shared/util/file-utils.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+        expect(parsedResult.length).toEqual(2);
+        expect(parsedResult[0][0]).toEqual('a');
+        expect(parsedResult[1][1]).toEqual('d');
+      });
+      it('with \r\n', () => {
+        const parsedResult = fileToArray('a\tb\nc\td', '\t');
+        expect(parsedResult.length).toEqual(2);
+        expect(parsedResult[0][0]).toEqual('a');
+        expect(parsedResult[1][1]).toEqual('d');
+      });
```
calvinlu3, line 23: Same test here, 2nd should use` \r\n`

src/test/javascript/spec/constants.ts
calvinlu3, line 1: We have a constants file at `src/test/javascript/constants.ts`.

calvinlu3 (approved):

calvinlu3:
@zhx828 Also the webdriver test is failing because it thinks the utils and constant files under `src/javascript/test/spec` are test files because in `wdio.conf.ts`, we have it looking `specs: ['./src/test/javascript/spec/**/*.ts'],`. Should probably move those files to `src/javascript/test/shared`?

zhx828:
> @zhx828 Originally Ben had it setup so that specs folder only contained tests, so the .spec suffix seemed redundant. However, if we want to includes other types of files in the specs folder, then we should update the config and also update test file names to include spec.

I will move to src/javascript/test/shared for now.

PR #361, opened 2024-06-18
Use google diff-match-patch to compare the text new/old content

zhx828:
## Tabs diff
Screenshot 2024-06-10 at 2 03 31 PM
Screenshot 2024-06-10 at 2 28 09 PM

## Stack diff 
Screenshot 2024-06-10 at 4 44 59 PM

## Merged diff
Screenshot 2024-06-10 at 4 24 15 PM

This fixes https://github.com/oncokb/oncokb-pipeline/issues/420

PR #368, opened 2024-06-20
Fix germline review

calvinlu3:
Missed germline changes in https://github.com/oncokb/oncokb-transcript/pull/361


zhx828 (approved):
Thank you!

PR #369, opened 2024-06-20
Add isGermline arg to all add and update firebase gene service methods

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/451

- [x] Fix production data


bprize15 (approved):
LGTM, just wondering why we are bumping 2 patch versions?

PR #372, opened 2024-06-26
Add additional e2e tests

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/366

After verifying that the tests work in GitHubActions/ Docker, I do plan on separating related test cases into their own files.


webpack/webpack.common.js
```diff
@@ -112,6 +112,7 @@ module.exports = async options => {
           },
           'Date.now': process.env.DOCKER ? `() => ${DEFAULT_DATE}` : 'Date.now',
           'new Date()': process.env.DOCKER ? `new Date(${DEFAULT_DATE})` : 'new Date()',
+          FIREBASE_DATA_EMULATOR_HOST: process.env.DOCKER ? 'firebase:9095' : 'localhost:9095',
```
calvinlu3, line 115: Based on Ben's comment, this env should go inside `"process.env"` obj.

calvinlu3: https://firebase.google.com/docs/emulator-suite/connect_firestore#admin_sdks

src/test/javascript/spec/e2e.ts
```diff
@@ -77,4 +91,39 @@ describe('Screenshot Tests', () => {
...
+    // Add a new mutation
+    const mutationNameInput = await $('input#add-mutation-modal-input');
+    await mutationNameInput.setValue('V600E');
+    await browser.keys('Enter');
+
+    const addMutationModal = await $('div[id="default-add-mutation-modal"]');
+    await addMutationModal.waitForDisplayed();
+
+    const modalConfirmBtn = await $('button=Confirm');
+    await modalConfirmBtn.click();
```
zhx828, line 112: We should share common actions. This one will be used repeatedly I imagine.

zhx828 (changes requested):

src/main/webapp/app/shared/button/AsyncSaveButton.tsx
```diff
@@ -16,7 +16,7 @@ export const AsyncSaveButton = ({ icon, isUpdate, confirmText, isSavePending, di
     return isUpdate ? 'Update' : 'Confirm';
   };
   return (
-    <Button disabled={disabled} color="primary" {...rest}>
+    <Button disabled={disabled} color="primary" id={'async-save-button'} {...rest}>
```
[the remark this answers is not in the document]
calvinlu3: That is a good point. Looks like we can use `useId` hook ([ref](https://react.dev/reference/react/useId#generating-unique-ids-for-accessibility-attributes:~:text=However%2C%20hardcoding%20IDs%20like%20this%20is%20not%20a%20good%20practice%20in%20React.%20A%20component%20may%20be%20rendered%20more%20than%20once%20on%20the%20page%E2%80%94but%20IDs%20have%20to%20be%20unique!%20Instead%20of%20hardcoding%20an%20ID%2C%20generate%20a%20unique%20ID%20with%20useId%3A)] from React to generate a unique id.

src/test/javascript/spec/e2e.ts
calvinlu3: Will create a test utils

package.json
```diff
@@ -93,6 +93,7 @@
     "dayjs": "1.11.11",
     "diff-match-patch": "^1.0.5",
     "firebase": "^10.11.1",
+    "firebase-admin": "^12.2.0",
```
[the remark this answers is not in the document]
calvinlu3: They are two separate libraries and both are using the latest (or near latest) versions.

src/main/webapp/app/shared/button/AsyncSaveButton.tsx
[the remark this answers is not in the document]
calvinlu3: Actually `useId` is only available in React v18 and we're on v17. I will just accept an `id` prop instead

src/main/webapp/app/service/firebase/firebase-gene-service.ts
```diff
@@ -392,8 +392,14 @@ export class FirebaseGeneService {
...
   ) => {
     const { hugoSymbol } = parseFirebaseGenePath(rctPath);
+
+    if (!uuid) {
+      uuid = generateUuid();
+    }
+
     if (initialUpdate) {
       return this.firebaseRepository.create(rctPath, newRelevantCancerTypes).then(() => {
+        this.firebaseRepository.create(`${rctPath}_uuid`, uuid);
```
bprize15, line 402: Can probably just bundle this in with the update

bprize15 (approved):
One minor comment, otherwise LGTM

src/main/webapp/app/service/firebase/firebase-gene-service.ts
calvinlu3: I should really also bundle the `_review` with the update too. I need to parse `rctPath` to get the Implication path in order to update. The `rctPath` given by the RCT Modal is something like `Genes/BRAF/mutations/0/tumors/0/TIs/0/treatments/0/excludedRCTs`, but updating would require the path to the treatment.

calvinlu3: I already made this change in my multi-location update PR, so I will leave the code for now if that's okay.

bprize15: Yeah for sure, makes sense to me

src/test/javascript/shared/utils.ts
```diff
@@ -0,0 +1,13 @@
+export const createMutationOnCurationPage = async (mutationName: string) => {
+  const mutationNameInput = await $('input#add-mutation-modal-input');
+  await mutationNameInput.setValue(mutationName);
```
zhx828, line 3: When adding mutation, we trigger an api call to annotate. Do we mock the data currently? Are  different mutation names using the same mock?

calvinlu3: Yeah they will all use the same mock since the API endpoint to annotate is the same (mutation name is passed in the request body). It may be possible to mock based on request body.

src/test/javascript/spec/e2e.ts
```diff
@@ -77,4 +92,30 @@ describe('Screenshot Tests', () => {
...
+
+    // Click to open mutation modal
+    const addMutationButton = await $('button=Add Mutation');
+    await addMutationButton.waitForDisplayed();
+    await addMutationButton.click();
+
+    await createMutationOnCurationPage('V600E');
+
+    // Delete the mutation on curation page
+    const mutationDeleteBtn = await $("div[data-testid='V600E-collapsible']").$("svg[data-icon='trash-can']");
```
zhx828, line 107: we need a testid generator to generate the id here. The knowledge of this id pattern is easy for us, but may not be to a new engineer without the context.

src/test/javascript/spec/screenshot.ts
```diff
@@ -116,4 +131,159 @@ describe('Screenshot Tests', () => {
...
+    await mutationEffectDesc.setValue('TP53 is the most frequently mutated gene in cancer. Additional text.');
+
+    // Go to review page to compare collapsible
+    const reviewButton = await $('button[data-testid="review-button"]');
+    await reviewButton.click();
+    const reviewCollapsible = await $('div[data-testid="collapsible"]');
+    await reviewCollapsible.waitForDisplayed();
+
+    const result = await browser.checkElement(reviewCollapsible, 'updated-gene-summary-review-collapsible', methodOptions);
+    expect(result).toBeLessThanOrEqual(ALLOWED_MISMATCH_PERCENTAGE);
```
zhx828, line 184: Is this gonna be our default for checking all screenshots? Maybe create our own expect method?

zhx828 (approved):
Looks good! Some minor comments added.

PR #373, opened 2024-06-26
Parse therapy uuid to name in review collapsible

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/463

### DIFF
--- a/src/main/webapp/app/pages/curation/collapsible/ReviewCollapsible.tsx
+++ b/src/main/webapp/app/pages/curation/collapsible/ReviewCollapsible.tsx
@@ -62,6 +62,8 @@ const ReviewCollapsibleBootstrapClass = {
   [ReviewAction.DEMOTE_MUTATION]: 'danger',
 };
 
+const TREATMENT_NAME_REGEX = /treatments\/\d+\/name$/;
+
 export interface IReviewCollapsibleProps {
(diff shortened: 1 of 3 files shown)


src/main/webapp/app/pages/curation/collapsible/ReviewCollapsible.tsx
```diff
@@ -311,8 +314,12 @@ export const ReviewCollapsible = ({
       );
     }
     if (reviewAction === ReviewAction.UPDATE || reviewAction === ReviewAction.NAME_CHANGE) {
-      const oldValue = reviewLevel.historyData.oldState as string;
-      const newValue = reviewLevel.historyData.newState as string;
+      let oldValue = reviewLevel.historyData.oldState as string;
+      let newValue = reviewLevel.historyData.newState as string;
+      if (reviewLevel.valuePath.includes('treatments')) {
+        oldValue = getTxName(drugList, oldValue);
```
bprize15, line 320: Let's say the new and old values are actually not the treatment name but the treatment level. In this case it looks like, `getTxName` will just do nothing. It does seem a bit misleading. Maybe we can use a regex here to say the valuePath ends with `treatments/\d`?

bprize15 (approved):
LGTM

PR #375, opened 2024-06-27
fix losing history when entity name changes

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/339


zhx828 (approved):
LGTM

PR #376, opened 2024-06-28
Updated backend model types

jfkonecn:


src/main/webapp/app/entities/companion-diagnostic-device/companion-diagnostic-device-update.tsx
```diff
@@ -47,7 +47,7 @@ export const CompanionDiagnosticDeviceUpdate = (props: ICompanionDiagnosticDevic
     for (const stv of specimenTypeValues) {
       if (!WHOLE_NUMBER_REGEX.test(stv.id)) {
         try {
-          const result = await props.createSpecimenType({ type: stv.id, name: stv.id });
+          const result = await props.createSpecimenType({ type: stv.id, name: stv.id, companionDiagnosticDevices: [] });
```
zhx828, line 50: When creating entity, if it is not associated with other entity, should this stay as null, instead of making a list?

src/main/webapp/app/shared/model/cancer-type.model.ts
```diff
@@ -3,18 +3,16 @@ import { IAssociation } from 'app/shared/model/association.model';
...
-  parent?: ICancerType | null;
-  associations?: IAssociation[] | null;
+  id: number;
+  code: string | null;
+  color: string | null;
+  level: number;
+  mainType: string;
+  subtype: string | null;
+  tissue: string | null;
+  tumorForm: TumorForm | undefined;
```
zhx828, line 13: This technically should not be undefined. Do you see otherwise?

src/main/webapp/app/shared/util/base-crud-store.ts
```diff
@@ -7,7 +7,7 @@ import { BaseReadStore } from 'app/shared/util/base-read-store';
 
 export abstract class BaseCrudStore<T extends object> extends BaseReadStore<T> {
   // TYPE-ISSUE: BaseStore needs a generic type
-  createEntity: ICrudPutAction<T> = this.updateHandler(this.create, 'Successfully created entity') as any;
+  createEntity: ICrudPutAction<Omit<T, 'id'>, T> = this.updateHandler(this.create, 'Successfully created entity') as any;
```
zhx828, line 10: 👍🏼

zhx828 (changes requested):
Thanks again for doing another massive PR. I didn't go through all entity models in details which I assume you followed the java model.

src/main/webapp/app/entities/article/article-update.tsx
```diff
@@ -69,8 +69,8 @@ export const ArticleUpdate = (props: IArticleUpdateProps) => {
           date: displayDefaultDateTime(),
         }
       : {
-          type: 'PUBMED',
           ...articleEntity,
+          type: articleEntity.type ?? 'PUBMED',
```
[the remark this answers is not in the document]
jfkonecn: Typescript now is assuming the `type` field is already populated. Even if the field is populate explicitly with null or undefined it'll override the default value so the assignment of `type` has to come after. If the field does not exist at all the previous code will still work, but that would violate the type definition.

src/main/webapp/app/shared/model/cancer-type.model.ts
jfkonecn: I wasn't sure what the default should be for `tumorForm` so I made [this change](https://github.com/oncokb/oncokb-transcript/pull/376/files#diff-46872c4ed9cdde649ee066ce4b1691794cf57530a281b543cefa0ae3da155385R639-R652) in [firebase-review-utils.tsx](https://github.com/oncokb/oncokb-transcript/blob/539c17494823f1ed786a5e72d260fb94ceb1c263/src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx#L639-L649) where I set it as undefined.

jfkonecn:
> Thanks again for doing another massive PR. I didn't go through all entity models in details which I assume you followed the java model.

I did. I just made all the fields required. The nullable values have to be explicitly set as null now.

src/main/webapp/app/shared/model/cancer-type.model.ts
zhx828: You can use TumorForm.mixed

jfkonecn: done

src/main/webapp/app/entities/companion-diagnostic-device/companion-diagnostic-device-update.tsx
jfkonecn: I'll make it null.

PR #377, opened 2024-06-28
allow searching by mutations

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/465

### DIFF
--- a/src/main/webapp/app/components/tabs/CurationHistoryTab.tsx
+++ b/src/main/webapp/app/components/tabs/CurationHistoryTab.tsx
@@ -15,6 +15,7 @@ import { IDrug } from 'app/shared/model/drug.model';
 import { GREY } from 'app/config/colors';
 import { FlattenedHistory, isBetweenDates } from 'app/shared/util/firebase/firebase-history-utils';
 import _ from 'lodash';
+import { compareMutationsDefault } from 'app/shared/util/firebase/firebase-utils';
 
 export interface ICurationHistoryTabProps extends StoreProps {
   historyData: FlattenedHistory[];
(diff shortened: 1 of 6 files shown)


src/main/webapp/app/components/tabs/CurationHistoryTab.tsx
```diff
@@ -50,6 +51,32 @@ const CurationHistoryTab = observer(({ historyData, getUsers, users, historyTabS
...
+    for (const mutation of mutations || []) {
+      for (const alteration of mutation.alterations || []) {
+        alts.add(alteration.alteration);
+      }
+      mutation.name.split(',').forEach(alt => alts.add(alt.trim()));
+    }
+    return alts;
+  }, [mutations]);
+
+  const mutationOptions = useMemo(() => {
```
zhx828, line 65: Should this be sorted? Do we use any default sorting for Author Select?

zhx828 (changes requested):
After selecting mutation and looking at history, I want to look at another mutation. But when I click on the select, the cursor is at the beginning of the previous search keyword I don't seem to be able to remove besides typing(I intuitively want to hit on the delete keystroke). Anyway we can make this more intuitive?

src/main/webapp/app/components/tabs/CurationHistoryTab.tsx
bprize15: There is not default sorting for the author select. However, I have changed the author dropdown to be sorted alphabetically.

I decided to sort the mutations using the default sort we came up with. However, we can use any of our mutation sorts for this order, so let me know if you prefer another sort.

bprize15:
> After selecting mutation and looking at history, I want to look at another mutation. But when I click on the select, the cursor is at the beginning of the previous search keyword I don't seem to be able to remove besides typing(I intuitively want to hit on the delete keystroke). Anyway we can make this more intuitive?



> After selecting mutation and looking at history, I want to look at another mutation. But when I click on the select, the cursor is at the beginning of the previous search keyword I don't seem to be able to remove besides typing(I intuitively want to hit on the delete keystroke). Anyway we can make this more intuitive?

I added some props do the `ReactSelect`. Now, backspace clears the select, and there is an `X` button to clear it as well.

src/main/webapp/app/components/tabs/CurationHistoryTab.tsx
zhx828: The default sorting is a bit odd without the context within the collapsible. But we can leave it as it until someone complains.

zhx828 (approved):
LGTM thanks.

PR #378, opened 2024-06-28
use oncokb icon as cursor when logo hovered

bprize15:
This PR is completely unnecessary, but I thought it could be a fun little easter egg for the curators to discover. There is no need for it to be merged whatsoever, but I figured I would let people take a look and see what they think.


zhx828:
😆 I don't mind!

src/main/webapp/oncogenic.svg
calvinlu3, line 1: Move to `src/main/webapp/content/images/oncogenic-black.svg`

zhx828 (approved):

PR #380, opened 2024-07-02
Explore new table with Excel-like filtering feature

calvinlu3:
This is related to https://github.com/oncokb/oncokb/issues/3739

I was unable to use ag-grid for oncokb-public because some of the packages are not compatible. We can let the curation team try it out first on the curation platform. If we find this library to be better, then we can explore upgrading packages for oncokb-public.

Here is a demo: https://www.ag-grid.com/example/

### Changes
- Converted gene list table to ag grid

Comparison:
image

The Excel style checkbox filters is only available in AG Grid Enterprise, but we can build our own sorting feature. Enterprise (paid version) will just help reduce development time.
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,8 @@
     "@sentry/react": "^7.112.2",
     "@sentry/webpack-plugin": "^2.17.0",
     "@superwf/mobx-react-router": "^7.4.0",
+    "ag-grid-community": "^32.0.0",
+    "ag-grid-react": "^32.0.0",
     "availity-reactstrap-validation": "^2.7.1",
     "axios": "1.6.8",
     "bootstrap": "5.3.3",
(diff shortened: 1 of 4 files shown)


calvinlu3:
After discussion:
1. We will not use ag-grid because the excel style filtering is locked behind enterprise license
2. We will explore free alternatives that have excel filtering 
3. If we cannot find a library, we want to see if we can move the cbioportal table out as a package and use it in our application. The cbioportal table already comes with checkbox filters and has searching.

calvinlu3:
**PrimeReact** seems like a good alternative (https://primereact.org/datatable/#filter)
- Has excel style filtering
- Good selection of ready to use UI components
- Can be integrated with Bootstrap styles

calvinlu3:
We are sticking with react-table and implementing our own filtering.

PR #382, opened 2024-07-09
Refactor to use multi location updates

calvinlu3:
fixes https://github.com/oncokb/oncokb-pipeline/issues/428

- [x] Would like to run screenshot tests after PR https://github.com/oncokb/oncokb-transcript/pull/372 is merged

### DIFF
--- a/src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
+++ b/src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
@@ -17,12 +17,11 @@ import {
   isCreateReview,
   isDeleteReview,
 } from '../../shared/util/firebase/firebase-review-utils';
-import { getFirebaseGenePath, getFirebaseVusPath } from '../../shared/util/firebase/firebase-utils';
+import { getFirebaseGenePath, getFirebaseMetaGenePath, getFirebaseVusPath } from '../../shared/util/firebase/firebase-utils';
 import { generateUuid, parseAlterationName } from '../../shared/util/utils';
 import { FirebaseVusService } from './firebase-vus-service';
 import { SentryError } from 'app/config/sentry-error';
 import { ActionType } from 'app/pages/curation/collapsible/ReviewCollapsible';
-import { GERMLINE_PATH } from 'app/config/constants/constants';
 
 export class FirebaseGeneReviewService {
   firebaseRepository: FirebaseRepository;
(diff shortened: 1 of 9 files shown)


calvinlu3:
Note that I only updated the firebase-gene-review-service methods to use multi-location updates even though we are dealing with number indexed arrays because - there should only be 1 person modifying a gene when it is under review (we have a ticket open to enforce this).

Once our firebase arrays all use firebase generated keys, then we can migrate firebase-gene-service to use multi-location updates as well.

src/main/webapp/app/shared/util/firebase/firebase-path-utils.ts
```diff
@@ -36,3 +36,12 @@ export const extractArrayPath = (valuePath: string) => {
...
   return { firebaseArrayPath, deleteIndex };
 };
+
+export const getImplicationPathFromRCTPath = (rctPath: string) => {
+  if (rctPath.endsWith('excludedRCTs')) {
+    const rctPathParts = rctPath.split('/');
+    rctPathParts.pop();
+    return rctPathParts.join('/');
+  }
+  return null;
```
zhx828, line 46: Can we add some comments here why we do this?

calvinlu3: The logic was part of initial refactoring effort, but unused now. I removed this method. Thanks!

zhx828 (approved):
LGTM, thanks!

src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
```diff
@@ -63,162 +73,174 @@ export class FirebaseGeneReviewService {
...
     }
 
     try {
-      await this.firebaseRepository.update(getFirebaseGenePath(isGermline, hugoSymbol), updateObject);
-      await this.firebaseMetaService.updateMeta(hugoSymbol, uuid, !isChangeReverted, isGermline);
+      await this.firebaseRepository.update('/', updateObject);
     } catch (error) {
-      throw new SentryError('Failed to update reviewable content', { firebasePath, currentValue, updateValue, review, uuid });
+      throw new SentryError('Failed to update reviewable content', updateObject);
     }
```
bprize15, line 90: Do you think we should return the `updateObject` even if we save? Feel like it would be nice to keep return value consistent

bprize15 (approved):
LGTM

PR #385, opened 2024-07-15
Added Save Gene Logic

jfkonecn:
closes oncokb/oncokb-pipeline#337

I don't have an authorization check for DEV_ROLE on the backend since I am using a proxy API to submit the core data and didn't want to block potential future requests. There is an authorization check for DEV_ROLE on the frontend though.

I moved the evidence logic into another folder to better organize the code so you can ignore changes inside `src/main/webapp/app/shared/util/core-evidence-submission/`


src/main/webapp/app/shared/button/SaveGeneButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+    setDisabled(true);
+    try {
+      const isGermline = false;
+      if (hugoSymbol === undefined) {
+        await firebaseGeneService.saveAllGenes(isGermline);
+      } else {
+        await firebaseGeneService.saveGene(isGermline, hugoSymbol);
+      }
+    } catch (e) {
+      console.error(e);
```
zhx828, line 24: We should also let user know the save failed.

src/main/webapp/app/shared/button/SaveGeneButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+    }
+    setDisabled(false);
+  }, [hugoSymbol]);
+  return (
+    <Button {...buttonProps} color="primary" outline onClick={onClickHandler} disabled={disabled}>
```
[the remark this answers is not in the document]
zhx828: We also need status indicating the save is complete.

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
```diff
@@ -0,0 +1,136 @@
+import _ from 'lodash';
+import { Drug, Gene, Mutation, Review, Treatment, Tumor, Vus, DrugCollection } from '../../model/firebase/firebase.model';
+
+export function getGeneData(geneData: Gene, excludeComments: boolean, onlyReviewedContent: boolean, drugList: DrugCollection): true | Gene {
```
zhx828, line 4: Should I review this file? I assume you refactored the code significantly?

src/main/webapp/app/shared/api/manual/drive-annotation-api.ts
```diff
@@ -0,0 +1,67 @@
+import { BASE_PATH, BaseAPI, RequestArgs } from '../generated/core/base';
+import globalAxios, { AxiosPromise, AxiosInstance, AxiosRequestConfig } from 'axios';
+import { Configuration } from '../generated/core';
+import { createRequestFunction } from '../generated/core/common';
+import { DUMMY_BASE_URL, assertParamExists, serializeDataIfNeeded, setSearchParams, toPathString } from '../generated/core/common';
+import { getDriveAnnotations as getDriveAnnotation } from 'app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission';
+
+export const DriveAnnotationAxiosParamCreator = function (configuration?: Configuration) {
```
zhx828, line 8: Should I review this file?

src/main/webapp/app/service/firebase/firebase-gene-service.ts
```diff
@@ -485,4 +497,46 @@ export class FirebaseGeneService {
   updateObject = async (path: string, value: any) => {
     await this.firebaseRepository.update(path, value);
   };
+
+  saveAllGenes = async (isGermlineProp: boolean) => {
+    const drugLookup = (await this.firebaseRepository.get(FB_COLLECTION.DRUGS)).val() as Record<string, Drug>;
+    const geneLookup =
+      ((await this.firebaseRepository.get(isGermlineProp ? FB_COLLECTION.GERMLINE_GENES : FB_COLLECTION.GENES)).val() as Record<
```
zhx828, line 504: The logic of getting gene collection based on isGermline should be moved to firebase utils. We do something similar in firebase-review-utils.tsx.

src/main/webapp/app/service/firebase/firebase-gene-service.ts
```diff
@@ -485,4 +497,46 @@ export class FirebaseGeneService {
...
+      >) ?? {};
+    for (const [hugoSymbol, gene] of Object.entries(geneLookup)) {
+      const maybeVus: Record<string, Vus> | null = vusLookup[hugoSymbol];
+      const args: Parameters<typeof getDriveAnnotations>[1] = {
+        gene,
+        vus: maybeVus == null ? undefined : Object.values(maybeVus),
+      };
+      const driveAnnotation = getDriveAnnotations(drugLookup, args);
+      // eslint-disable-next-line no-console
+      console.log(driveAnnotation);
```
zhx828, line 521: Is this intentional?

zhx828 (changes requested):

README.md
```diff
@@ -202,6 +202,7 @@
        (user_id, authority_name)
```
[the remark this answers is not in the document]
jfkonecn: This is add a new user not to update the authority list.

src/main/webapp/app/shared/api/manual/drive-annotation-api.ts
jfkonecn: yeah

src/main/webapp/app/service/firebase/firebase-gene-service.ts
jfkonecn: nope, removing...

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
jfkonecn: Any inside of `src/main/webapp/app/shared/util/core-drive-annotation-submission` is new code. Anything in `src/main/webapp/app/shared/util/core-evidence-submission` is old code, but was move to that folder for organizational purposes.

src/main/webapp/app/shared/button/SaveGeneButton.tsx
jfkonecn: I didn't think it matter since it was for developers, but I think you're right. I'll add a toast message.

src/main/webapp/app/service/firebase/firebase-gene-service.ts
jfkonecn: I see there's a `getGenePathFromValuePath` function, but it does not have logic to actually fetch.
Did you want me to use that function?

zhx828: I think this one is just to see which gene collection to use.

README.md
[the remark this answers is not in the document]
zhx828: Ah yeah, sorry. My bad.

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
```diff
@@ -0,0 +1,136 @@
...
+      const reviewKey = key + '_review';
+      const maybeReview = data[reviewKey];
+      if (onlyReviewedContent && typeof maybeReview === 'object' && maybeReview !== null && 'lastReviewed' in maybeReview) {
+        data[key] = maybeReview.lastReviewed;
+      }
+    }
+  }
+}
+
+export function getVUSData(vus: Vus[], excludeComments: boolean) {
```
zhx828, line 82: We haven't needed to include comments for parsing in server side for a long time. Let's remove.

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
zhx828, line 4: Should we have all functions exposed and add few unit tests for each? Or you think the `getGeneData` and `getVUSData` covered them pretty well?

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
```diff
@@ -0,0 +1,136 @@
...
+    processData(mutation.mutation_effect, ['oncogenic', 'effect', 'description'], excludeComments, onlyReviewedContent);
+    for (const tumor of mutation.tumors ?? []) {
+      if (shouldExclude(onlyReviewedContent, tumor.cancerTypes_review)) {
+        tempTumors.push(tumor);
+        continue;
+      }
+      // process tumor cancerTypes
+      processData(tumor, ['summary', 'diagnosticSummary', 'prognosticSummary'], excludeComments, onlyReviewedContent);
+      processData(tumor.diagnostic, ['level', 'description', 'relevantCancerTypes'], excludeComments, onlyReviewedContent);
+      processData(tumor.prognostic, ['level', 'description', 'relevantCancerTypes'], excludeComments, onlyReviewedContent);
```
zhx828, line 25: @bprize15  could you remind me whether we still use relevantCancerTypes or we started to use excludedCancerTypes?

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
```diff
@@ -0,0 +1,136 @@
...
+        const tempTreatments: TempTreatment[] = [];
+        for (const treatment of ti.treatments ?? []) {
+          if (shouldExclude(onlyReviewedContent, treatment.name_review)) {
+            tempTreatments.push(treatment);
+            return true;
+          }
+          (treatment as TempTreatment).name = drugUuidToDrug(treatment.name, drugList);
+          processData(
+            treatment,
+            ['level', 'propagation', 'propagationLiquid', 'indication', 'description'],
```
zhx828, line 38: Is fdaLevel not here?

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
zhx828: My apology, could've saved you some time for writing these tests.

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
```diff
@@ -0,0 +1,136 @@
...
+  for (const vusItem of vusData) {
+    if (excludeComments) {
+      delete vusItem.name_comments;
+    }
+    vusDataArray.push(vusItem);
+  }
+  return vusDataArray;
+}
+
+function shouldExclude(onlyReviewedContent: boolean, reviewObj: Review | undefined) {
```
zhx828, line 95: @calvinlu3 need your input here to check whether this still makes sense.

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
```diff
@@ -0,0 +1,136 @@
...
+    return getDrugsByUuids(keys, drugList);
+  } else {
+    return {};
+  }
+}
+
+function therapyStrToArr(key: string) {
+  if (key) {
+    return key.split(',').map(function (element) {
+      return element.trim().split(' + ');
```
zhx828, line 111: let's split by `+` then trim on the safe side.

zhx828 (changes requested):
Thank you for writing all the tests and there are a lot of tests to write. And yes, we need them!

src/main/webapp/app/service/firebase/firebase-gene-service.ts
```diff
@@ -485,4 +497,46 @@ export class FirebaseGeneService {
...
+        string,
+        Gene
+      >) ?? {};
+    const vusLookup =
+      ((await this.firebaseRepository.get(isGermlineProp ? FB_COLLECTION.GERMLINE_VUS : FB_COLLECTION.VUS)).val() as Record<
+        string,
+        Record<string, Vus>
+      >) ?? {};
+    for (const [hugoSymbol, gene] of Object.entries(geneLookup)) {
+      const maybeVus: Record<string, Vus> | null = vusLookup[hugoSymbol];
```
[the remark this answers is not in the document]
jfkonecn: `hasVus` sounds like it should be a `true`/`false` value to me. What about `optionalVus`?

src/main/webapp/app/shared/util/core-evidence-submission/core-evidence-submission.spec.ts
```diff
@@ -0,0 +1,746 @@
...
+          valuePath: 'mutations/0/mutation_effect/oncogenic',
+          updateTime: getTimeFromDateString('2000-01-01'),
+          gene: createMockGene({
+            name: hugoSymbol,
+            mutations: [
+              createMockMutation({
+                mutation_effect: createMockMutationEffect({
+                  description: 'Mutation Effect',
+                  effect_uuid: 'a7046e8d-af62-4284-a14d-fc145134529a',
+                  oncogenic: FIREBASE_ONCOGENICITY.LIKELY,
```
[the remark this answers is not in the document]
jfkonecn: The type on mutation effect is `FIREBASE_ONCOGENICITY`

```ts
export class MutationEffect {
  description = '';
  description_review?: Review;
  description_uuid: string = generateUuid();
  effect = '';
  effect_review?: Review;
  effect_uuid: string = generateUuid();
  oncogenic: FIREBASE_ONCOGENICITY | '' = '';
  oncogenic_review?: Review;
  oncogenic_uuid: string = generateUuid();
  short = '';

  // Germline
  pathogenic: `${PATHOGENICITY}` | '' = '';
  pathogenic_review?: Review;
  pathogenic_uuid: string = generateUuid();
}
```

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
bprize15: If I remember correctly, we are talking about how a curator might exclude a cancer type from a treatment?

If so, I believe we decided that a curator might used the RCT modal when they need to exclude many cancer types and that they would most likely exclude the cancer type directly in the name (using `excludedCancerTypes`). 

Since we support both ways, I imagine the logic would need to consider both methods of excluding cancer types. However, based on our conversations (among ourselves and with curators), I believe they most commonly use `excludedCancerTypes`.

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
jfkonecn: no worries

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
jfkonecn: My preference is to just test the public functions. If you start adding test every single function it become more difficult to refactor since you'll break a lot more of tests just because you refactored not because you wrote bugs.

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
jfkonecn: It's copied from the [old code](https://github.com/oncokb/curation-platform/blob/master/app/scripts/services/mainutils.js#L481-L488)

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
jfkonecn: @zhx828 should I change `relevantCancerTypes` to `excludedCancerTypes`?

src/main/webapp/app/shared/util/core-evidence-submission/core-evidence-submission.spec.ts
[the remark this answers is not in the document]
zhx828: @jfkonecn yeah but for `oncogenic` which is correct. For `effect`, it should be MUTATION_EFFECT | ''

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
zhx828: @jfkonecn let's pass both in to the server side. We might have legacy code storing relevantCancerTypes.

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
jfkonecn: It only impacts if the last reviewed data is used, but my test don't break if I change it. I'm going to adjust my tests so they will break if I make changes to this logic though. I'll also add `fdaLevel` to the array.

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
zhx828: @calvinlu3 ☝🏼

calvinlu3: Is there a use case for when `onlyReviewedContent` is false? If we're saving the data, we are currently skipping any data that still requires review.

I do believe that we still need a check to determine if `field_review` has a review object and the review object has any of the properties: `added, removed, demotedToVus, promotedToMutation, initialUpdate`

zhx828: @calvinlu3 we used to have a function to download a PDF which uses the latest data not reviewed content. That said, if we to implement the same functionality, I would download the review data instead. I can't think of any use case which onlyReviewedContent is false.

calvinlu3: That make sense then. We should probably keep this logic then in case we need to query for all (including items pending review).

However I would modify the condition to be
```
reviewObj && ((onlyReviewedContent && reviewObj.added === true && reviewObj.promotedToMutation === true && reviewObj.initialUpdate === true) ||
```
those are two new properties added to Review object that when true, also means that the content has not been reviewed yet and should be excluded

PR #386, opened 2024-07-18
Add beta site banner

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/475

- Add beta site banner
- Close sidebar by default on gene list page

image


src/main/webapp/app/shared/userMessage/BetaSiteMessage.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import * as styles from './styles.module.scss';
+import { Container } from 'reactstrap';
+import { BANNER_HEIGHT } from 'app/stores/layout.store';
+import { IRootStore } from 'app/stores';
+import { componentInject } from '../util/typed-inject';
+import { observer } from 'mobx-react';
+
+const BetaSiteMessage = (props: StoreProps) => {
```
zhx828, line 9: Should this component be called `Banner` and have the message as pass in component which is BetaSiteMessage?

zhx828 (changes requested):

src/main/webapp/app/shared/userMessage/BetaSiteMessage.tsx
calvinlu3: Thought about making it general, but seems like this is the only use case right now. Called it specifically BetaSiteMessage because it has some logic to check if is beta site

zhx828: Sounds good. Good to merge if you could take a look at the test.

zhx828 (approved):
LGTM

PR #388, opened 2024-07-19
Check if mutations data is available before using

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/481

### DIFF
--- a/src/main/webapp/app/shared/table/GenomicIndicatorsTable.tsx
+++ b/src/main/webapp/app/shared/table/GenomicIndicatorsTable.tsx
@@ -19,6 +19,7 @@ import { getHexColorWithAlpha } from '../util/utils';
 import { parseFirebaseGenePath } from '../util/firebase/firebase-path-utils';
 import GenomicIndicatorsHeader from 'app/pages/curation/header/GenomicIndicatorsHeader';
 import { SentryError } from 'app/config/sentry-error';
+import _ from 'lodash';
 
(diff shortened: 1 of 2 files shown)


src/main/webapp/app/shared/util/firebase/firebase-utils.tsx
```diff
@@ -51,7 +51,7 @@ export const getAlterationName = (alteration: Alteration) => {
   }
   return '';
 };
-export const getMutationName = (name: string, alterations: Alteration[]) => {
+export const getMutationName = (name: string | undefined, alterations: Alteration[] | null | undefined) => {
```
calvinlu3, line 54: Types copied from John's strict null check PR.

src/main/webapp/app/shared/table/GenomicIndicatorsTable.tsx
```diff
@@ -19,6 +19,7 @@ import { getHexColorWithAlpha } from '../util/utils';
 import { parseFirebaseGenePath } from '../util/firebase/firebase-path-utils';
 import GenomicIndicatorsHeader from 'app/pages/curation/header/GenomicIndicatorsHeader';
 import { SentryError } from 'app/config/sentry-error';
+import _ from 'lodash';
```
bprize15, line 22: Looks like this is unnecessary

bprize15 (approved):
One small comment, otherwise LGTM

PR #389, opened 2024-07-22
Add Pathogenic Variants to Mutations list when adding genomic indicator

zhx828:
The genomic indicator is automatically associated with Pathogenic Variants. We like to make sure it's included in the mutation list.

This fixes https://github.com/oncokb/oncokb-pipeline/issues/389


bprize15 (approved):
LGTM

PR #390, opened 2024-07-25
Curation platform supports 3 letter code

carrotcatnyc:
issue #446 (Curation platform supports 3 letter code)

### DIFF
--- a/src/main/webapp/app/shared/modal/AddMutationModal.tsx
+++ b/src/main/webapp/app/shared/modal/AddMutationModal.tsx
@@ -281,7 +281,7 @@ function AddMutationModal({
     variantName?: string,
   ): AlterationData {
     const alteration = entityStatusAlteration.entity;
-    return {
+    const alterationData: AlterationData = {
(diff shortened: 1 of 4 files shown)


src/main/webapp/app/shared/modal/AddMutationModal.tsx
```diff
@@ -297,6 +297,13 @@ function AddMutationModal({
       warning: entityStatusAlteration.warning ? entityStatusAlteration.message : null,
       error: entityStatusAlteration.error ? entityStatusAlteration.message : null,
     };
+
+    // if the backend's response is different from the frontend response, set them equal to each other.
+    if (alteration?.alteration !== alterationName) {
```
zhx828, line 302: I believe alterationName is computed with other fields within the alteration model. For instance, if we have comment on the model, alterationName includes it. This change here will overwrite it.

zhx828 (changes requested):

zhx828 (approved):

PR #393, opened 2024-07-26
Added name changes to review logic

jfkonecn:


src/main/webapp/app/shared/util/core-evidence-submission/core-evidence-submission.ts
```diff
@@ -85,6 +89,12 @@ export function pathToGetEvidenceArgs({
     type = EvidenceEvidenceTypeEnum.DiagnosticImplication;
   } else if (/^mutations\/\d+\/mutation_effect\/oncogenic/.test(valuePath)) {
     type = EvidenceEvidenceTypeEnum.Oncogenic;
+  } else if (/^mutations\/\d+\/name/.test(valuePath)) {
+    type = 'MUTATION_NAME_CHANGE';
+  } else if (/^mutations\/\d+\/tumors\/\d+\/cancerTypes/.test(valuePath)) {
+    type = 'TUMOR_NAME_CHANGE';
```
zhx828, line 95: We also now need to monitor `excludedCancerTypes`

src/main/webapp/app/shared/util/core-evidence-submission/core-evidence-submission-utils.ts
```diff
@@ -101,3 +101,148 @@ export function getNewPriorities(list: Treatment[], unapprovedUuids: string | st
...
+    }
+  | {
+      type: 'treatment';
+      obj: Treatment;
+    }
+) & {
+  uuids?: string[];
+  uuidType?: 'insideOnly' | 'evidenceOnly' | 'sectionOnly';
+};
+export function collectUUIDs({ type, obj, uuids = [], uuidType }: CollectUUIDsArgs): string[] {
```
zhx828, line 126: Do you like how this method is written? We can also break them up to collectMutationUuids etc.

zhx828 (changes requested):

src/main/webapp/app/shared/util/core-evidence-submission/core-evidence-submission-utils.ts
jfkonecn: I like it. This kind of style is becoming more and more popular. I will admit it looks better when you have [pattern matching support](https://dev.java/learn/pattern-matching/) built into the language.

zhx828: I was ahead of JAVA 16 :) Could you also move the comment from the legacy to here as well? I think they are pretty useful. And update it accordingly.

src/main/webapp/app/shared/util/core-evidence-submission/type-specific-resolvers.ts
```diff
@@ -90,6 +91,26 @@ function handleGeneSummary({ gene, evidenceData }: Pick<GetEvidenceArgs, 'gene'>
...
+  evidenceData.data.excludedCancerTypes = tumor.excludedCancerTypes as TumorType[];
+}
+
+function handleTreatmentNameChange({
+  evidenceData,
+  treatment,
+}: Pick<GetEvidenceArgs, 'treatment'> & { evidenceData: InitializeEvidenceDataRtn }) {
+  evidenceData.data.excludedCancerTypes = treatment.excludedRCTs as TumorType[];
+}
+
```
zhx828, line 113: Talking about pattern matching, do you want to merge these three?

zhx828 (commented):
All the core changes we made so far are for the somatic. We will incorporate germline soon. Could you think about how to design these methods so it would be easier when we  start to incorporate germline?
A big background in case you are not so familiar with somatic/germline structure. They have been separated through firebase structure. We are fetching gene info from Genes collection or Germline_Genes collection based on somatic/germlien status. They share the same Gene/Mutation/etc firebase model though which means we will collect different uuid based on the the status.

src/main/webapp/app/shared/util/core-evidence-submission/type-specific-resolvers.ts
jfkonecn: The reason why didn't do that here is because I have dedicated functions for every case in the switch statement so I thought I would be less confusing if I wrote it this way.

zhx828 (approved):
LGTM, thanks!

PR #394, opened 2024-07-29
Fix cancer type change reverted check

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/488

- [ ] Cleanup production firebase's meta collection

### Issues
1. There may be scenario where `excludedCancerTypes` is not defined yet but since `cancerTypes` was modified, only `cancerTypes_uuid` was added to meta to mark as review. Before that change has been reviewed, making a change to `excludedCancerTypes` will cause `cancerTypes_uuid, excludedCancerTypes_uuid` to be added to meta. After accepting/rejecting change, only `cancerTypes_uuid, excludedCancerTypes_uuid` was removed. The fix is to find all meta keys where either uuids are present and remove upon review.
2. Cancer types are not properly compared causing `isChangeReverted` to not be accurate.

### DIFF
--- a/src/main/webapp/app/shared/modal/modify-cancer-type-modal.store.ts
+++ b/src/main/webapp/app/shared/modal/modify-cancer-type-modal.store.ts
@@ -78,13 +78,13 @@ export class ModifyCancerTypeModalStore {
   setIncludedCancerTypes(cancerTypes: CancerTypeSelectOption[], allCancerTypes: Tumor[], cancerTypeToEditUuid?: string) {
     this.setIsErrorIncludedAndExcluded(cancerTypes, this.excludedCancerTypes);
     this.setIsErrorDuplicate(cancerTypes, this.excludedCancerTypes, allCancerTypes, cancerTypeToEditUuid);
-    this.includedCancerTypes = cancerTypes.map(removeCancerTypeCode);
+    this.includedCancerTypes = cancerTypes;
   }
 
   setExcludedCancerTypes(cancerTypes: CancerTypeSelectOption[], allCancerTypes: Tumor[], cancerTypeToEditUuid?: string) {
     this.setIsErrorIncludedAndExcluded(this.includedCancerTypes, cancerTypes);
     this.setIsErrorDuplicate(this.includedCancerTypes, cancerTypes, allCancerTypes, cancerTypeToEditUuid);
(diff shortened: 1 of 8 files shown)


src/main/webapp/app/shared/modal/modify-cancer-type-modal.store.ts
```diff
@@ -109,3 +109,13 @@ export class ModifyCancerTypeModalStore {
     this.isRetryButtonClicked = false;
   }
 }
+
+// Once https://github.com/oncokb/oncokb-pipeline/issues/413 is fixed, we can remove this function.
+// Then we should complete this ticket: https://github.com/oncokb/oncokb-pipeline/issues/494 to backfill
+// all place where code == "" in firebase.
+const removeCancerTypeCode = (cancerTypeOption: CancerTypeSelectOption) => {
+  if (cancerTypeOption.level < 0) {
+    cancerTypeOption.code = null;
```
zhx828, line 118: Do we really need to do this?

src/main/webapp/app/shared/util/firebase/firebase-utils.tsx
```diff
@@ -877,3 +878,22 @@ export function findNestedUuids(obj: any, uuids: string[] = []) {
 
   return uuids;
 }
+
+export function areCancerTypeArraysEqual(a: CancerType[], b: CancerType[]) {
```
zhx828, line 882: are these two sorted list?

src/main/webapp/app/shared/modal/modify-cancer-type-modal.store.ts
calvinlu3: Before that ticket is resolved, yes.
Currently for data release, I have to manually remove all code for level < 0.

src/main/webapp/app/shared/util/firebase/firebase-utils.tsx
calvinlu3: They are not sorted, so if ordering changes, then this tumor name will appear in review mode.

zhx828:
@calvinlu3 let's update the PR based on the latest rc.

zhx828 (approved):
LGTM, thanks!

PR #395, opened 2024-07-30
Support slash in review collapsible title

calvinlu3:
This fixes https://github.com/oncokb/oncokb-pipeline/issues/490

### Changes
- Replace slash in title used for separating sections to double right arrow to distinguish from a section have a slash in their name

image

image


src/main/webapp/app/pages/curation/collapsible/ReviewCollapsible.tsx
```diff
@@ -432,10 +428,10 @@ export const ReviewCollapsible = ({
 
   return (
     <Collapsible
-      idPrefix={baseReviewLevel.title}
+      idPrefix={baseReviewLevel.titleParts.join('/')}
```
zhx828, line 431: using the slash here as id will be ok?

calvinlu3: Slash here is fine because we don't do any splitting on idPrefix

zhx828 (approved):

PR #398, opened 2024-07-31
Can now delete evidences

jfkonecn:


src/main/webapp/app/shared/api/manual/evidence-api.ts
```diff
@@ -1,16 +1,17 @@
-import { getEvidence } from 'app/shared/util/core-evidence-submission/core-evidence-submission';
```
zhx828, line 1: I forgot to ask in the previous PRs. Why don't we automatically generate these clients?

zhx828 (approved):

src/main/webapp/app/shared/api/manual/evidence-api.ts
jfkonecn: I believe it was because the legacy API isn't on the swagger pages.

PR #400, opened 2024-08-01
Add async to getSessionGen

calvinlu3:
@zhx828 @bprize15 
In `getSessionGen`, if the `axios.get('/api/account')` throws a 401 error, it never reaches the catch block. John and I found out that if you add `.catch` to axios.get(...) OR add async to generator function, the exception ends up being caught.

Let us know if you might know what the cause is.

### DIFF
--- a/src/main/webapp/app/stores/authentication.store.ts
+++ b/src/main/webapp/app/stores/authentication.store.ts
@@ -102,7 +102,7 @@ export class AuthStore extends BaseStore {
     }
   }
 
-  *getSessionGen() {
+  async *getSessionGen() {
     if (this.isAuthenticated) {
       return this.account;
     }


bprize15:
> @zhx828 @bprize15 In `getSessionGen`, if the `axios.get('/api/account')` throws a 401 error, it never reaches the catch block. John and I found out that if you add `.catch` to axios.get(...) OR add async to generator function, the exception ends up being caught.
> 
> Let us know if you might know what the cause is.

I assume you are seeing the exception not get caught because the function is not asynchronous without the async keyword. This means that the call to `yield` will not pause execution. This means that by the time the exception is thrown, we have already exited the `try/catch`.

bprize15 (approved):
LGTM

zhx828 (approved):

PR #402, opened 2024-08-06
add vus with special interpretation option for germline

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/501


calvinlu3 (approved):
LGTM

PR #405, opened 2024-08-08
Fix cancer type select

calvinlu3:
fixes https://github.com/oncokb/oncokb-pipeline/issues/510

### DIFF
--- a/src/main/webapp/app/shared/select/CancerTypeSelect.tsx
+++ b/src/main/webapp/app/shared/select/CancerTypeSelect.tsx
@@ -14,15 +14,17 @@ interface ICancerTypeSelectProps<IsMulti extends boolean> extends SelectProps<Ca
   disabledOptions?: readonly CancerTypeSelectOption[];
 }
 
-export type CancerTypeSelectOption = {
-  label: string;
-  value: number;
-  code: string;
-  mainType: string;
-  subtype: string;
-  level: number;
-  isDisabled?: boolean;
-};
+export type CancerTypeSelectOption =
+  | {
+      label: string;
+      value: number;
+      code: string;
+      mainType: string;
+      subtype: string;
+      level: number;
+      isDisabled?: boolean;
+    }
+  | undefined;
 
 const getAllMainTypes = (cancerTypeList: ICancerType[]) => {
   return _.uniq(cancerTypeList.filter(cancerType => cancerType.level && cancerType.level <= 0)).sort();
@@ -120,7 +122,9 @@ const CancerTypeSelect = <IsMulti extends boolean>(props: ICancerTypeSelectProps
       additional={{ ...defaultAdditional, type: SearchOptionType.CANCER_TYPE }}
       loadOptions={loadCancerTypeOptions}
       reduceOptions={reduceGroupedOptions}
-      isOptionDisabled={option => disabledOptions?.some(disabled => disabled.value === option.value) || false}
+      isOptionDisabled={option =>
+        disabledOptions?.some(disabled => disabled !== undefined && option !== undefined && disabled.value === option.value) || false
+      }
       cacheUniqs={[props.value]}
       placeholder="Select a cancer type..."
       isClearable
--- a/src/main/webapp/app/shared/select/CancerTypeSelect.tsx
+++ b/src/main/webapp/app/shared/select/CancerTypeSelect.tsx
@@ -71,7 +71,7 @@ const CancerTypeSelect = <IsMulti extends boolean>(props: ICancerTypeSelectProps
     CancerTypeSelectOption,
     GroupBase<CancerTypeSelectOption>,
     { page: number; type: SearchOptionType }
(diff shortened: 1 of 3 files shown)


src/main/webapp/app/shared/select/CancerTypeSelect.tsx
```diff
@@ -92,8 +92,16 @@ const CancerTypeSelect = <IsMulti extends boolean>(props: ICancerTypeSelectProps
     if (searchWord) {
       // Since options are cached by react-select-async-paginate, we only include options that are
       // not present in prevOptions to avoid duplicates.
-      options[0].options = _.differenceBy(options[0].options, 'options' in prevOptions[0] ? prevOptions[0]?.options ?? [] : [], 'label');
```
calvinlu3, line 95: The code stop executing here because `prevOption[0]` is undefined and you can use the `in` operator on undefined. @jfkonecn Is there a rule we can add so that we will get an error here? Or is it because prevOption[0]'s is not defined as possibly undefined?

jfkonecn: It's because it's not defined as undefined. For all I/O (axios call in this case), as far as I'm aware, typescript assumes the type you give it is correct. To be absolutely sure the types are correct you have to assume the types are `unknown` and then check the type from there. It's pain to do so most people don't do it. You can use [zod](https://zod.dev/), but you will have to pay a price on performance. 

Here's an example with `fetch`.
[The following code](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAKg9gETgcTnAJlAvFA3gKCiKgFcBnCAJwEl0AuKAOxIFsAjKgGkOIEt6mrDpW7EowXsAA2EBmWCVejAOajiAYzgswM4BAFs0MgIaMA3PgC++fMbIhG6qADMSjiXEZRlEYPCSoGAAU-AzM7FQAlAwACpRavBQAPP4oaOgAfHg8RJqM8lCUEGRgnhTYUMYA7saSLr7qABZBAAaNwMBgZHQA9D0AVmSeOsbqEI1wUuhUAHSgYLya0zOaLD3AGHBkPQAkuPyWLZEWYn1QvNpSi3WmIFA5UHkFG0gV1bXAhcWl+RAzg54gscHmcLjprp8XnBKmRYIg0hgHkVgCRKF4oRZrPgzoxoYkyCRivgfH54YF0EEAIyRB5zRoQRhBAAe2CyT0mfykcGUzLmkhkkRpYhWxmATWZrMeZQ5Myo8UoQQAREMWL5GkplFAqgzPlV4ipFZwoEzBRYbPNoKkAELGTA4AhichUWgMXDnAThYSWADaAF01ERQoIIiIHhJpLIoPJFCoA1LLr59AxDBzTJibHYHE5XO5eJ5vL5rbaQh6hFFYvEWIkICl4TbMtkxE9PkUSmVoDh3nVnA1mm0Ol1egMhowRmMJlNZvNFhg-qt1pttnsDkcTsRQZcIZVGHcHs3xPC3jU6q2fhR-iOgWuiBvweo6lCYXCkPWkb5Uej4dexPcsWdgI08RVLCcpwJQxKFnWxbUrSAEMhKWBstKMgzFyPJMjMTo0Og3oAAy+jM-CCrS6iiuKLKIVK+QyqBCrKloaoalqOpavqyiGsappAA) will have no type errors, but will break at run time because the types are wrong.

```ts
type ToDoGood = {
    userId: number,
    id: number,
    title: string,
    completed: boolean;
}

async function getToDoGood(id: number): Promise<ToDoGood> {
    const response = await fetch(`https://jsonplaceholder.typicode.com/todos/${id}`);
    // implicit any 
    const toDo = await response.json();
    // implicit toDo as ToDoGood
    return toDo;
}

// no issues
getToDoGood(1)
    .then(x => console.log(x.title))
    .catch(x => console.error("something went wrong", x));


type ToDoBad = {
    userId: { id: number}[],
    id: number,
    title: string,
    completed: boolean;
}

async function getToDoBad(id: number): Promise<ToDoBad> {
    const response = await fetch(`https://jsonplaceholder.typicode.com/todos/${id}`);
    // implicit any 
    const toDo = await response.json();
    // implicit toDo as ToDoBad
    return toDo;
        
}

// throws error
getToDoBad(1)
    .then(x => console.log(x.userId[0].id))
    .catch(x => console.error("something went wrong", x));
```

jfkonecn: This should give you type errors if you use `in` without the undefined check.

```ts
  const loadCancerTypeOptions: LoadOptions<
    CancerTypeSelectOption | undefined,
    GroupBase<CancerTypeSelectOption | undefined>,
    { page: number; type: SearchOptionType }
  > = async (searchWord, prevOptions, { page, type } = { page: 0, type: SearchOptionType.CDX }) => {
```

jfkonecn: @zhx828 looks like this error happened because of my strict null check change.

jfkonecn (approved):

PR #408, opened 2024-08-09
Split web driver e2e and screenshots action into two jobs

zhx828:


jfkonecn:
In order to speed up the build you might be able to use the [COPY](https://docs.docker.com/reference/dockerfile/#copy) command. They have an example using go that looks like this:
```dockerfile
FROM golang AS build
WORKDIR /app
RUN --mount=type=bind,target=. go build -o /myapp ./cmd

COPY --from=build /myapp /usr/bin/
```
So for example in the app docker file, you could do something like this where you add a final step that just copies the artifacts over:

```dockerfile
FROM node:20.12.2 as base

RUN apt-get update

RUN apt-get install build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev libnss3 -y


FROM base as node_modules

RUN mkdir /app
WORKDIR /app

COPY package*.json .
COPY yarn.lock .

RUN yarn install


FROM node_modules as build
WORKDIR /app
COPY . .
RUN yarn run webapp:build:dev

FROM node:20.12.2
WORKDIR /app
COPY --from=build /app/target .
```

jfkonecn (changes requested):
In order to speed up the build you might be able to use the [COPY](https://docs.docker.com/reference/dockerfile/#copy) command. They have an example using go that looks like this:
```dockerfile
FROM golang AS build
WORKDIR /app
RUN --mount=type=bind,target=. go build -o /myapp ./cmd

COPY --from=build /myapp /usr/bin/
```
So for example in the app docker file, you could do something like this where you add a final step that just copies the artifacts over:

```dockerfile
FROM node:20.12.2 as base

RUN apt-get update

RUN apt-get install build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev libnss3 -y


FROM base as node_modules

RUN mkdir /app
WORKDIR /app

COPY package*.json .
COPY yarn.lock .

RUN yarn install


FROM node_modules as build
WORKDIR /app
COPY . .
RUN yarn run webapp:build:dev

FROM node:20.12.2
WORKDIR /app
COPY --from=build /app/target .
```

bprize15:
> In order to speed up the build you might be able to use the [COPY](https://docs.docker.com/reference/dockerfile/#copy) command. They have an example using go that looks like this:
> 
> ```dockerfile
> FROM golang AS build
> WORKDIR /app
> RUN --mount=type=bind,target=. go build -o /myapp ./cmd
> 
> COPY --from=build /myapp /usr/bin/
> ```
> 
> So for example in the app docker file, you could do something like this where you add a final step that just copies the artifacts over:
> 
> ```dockerfile
> FROM node:20.12.2 as base
> 
> RUN apt-get update
> 
> RUN apt-get install build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev libnss3 -y
> 
> 
> FROM base as node_modules
> 
> RUN mkdir /app
> WORKDIR /app
> 
> COPY package*.json .
> COPY yarn.lock .
> 
> RUN yarn install
> 
> 
> FROM node_modules as build
> WORKDIR /app
> COPY . .
> RUN yarn run webapp:build:dev
> 
> FROM node:20.12.2
> WORKDIR /app
> COPY --from=build /app/target .
> ```

I like this solution in general, but I don't think GitHub can currently take advantage of multi-stage builds because it always busts the build cache. 

> In order to speed up the build you might be able to use the [COPY](https://docs.docker.com/reference/dockerfile/#copy) command. They have an example using go that looks like this:
> 
> ```dockerfile
> FROM golang AS build
> WORKDIR /app
> RUN --mount=type=bind,target=. go build -o /myapp ./cmd
> 
> COPY --from=build /myapp /usr/bin/
> ```
> 
> So for example in the app docker file, you could do something like this where you add a final step that just copies the artifacts over:
> 
> ```dockerfile
> FROM node:20.12.2 as base
> 
> RUN apt-get update
> 
> RUN apt-get install build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev libnss3 -y
> 
> 
> FROM base as node_modules
> 
> RUN mkdir /app
> WORKDIR /app
> 
> COPY package*.json .
> COPY yarn.lock .
> 
> RUN yarn install
> 
> 
> FROM node_modules as build
> WORKDIR /app
> COPY . .
> RUN yarn run webapp:build:dev
> 
> FROM node:20.12.2
> WORKDIR /app
> COPY --from=build /app/target .
> ```

I agree with this optimization, but I also think we should look into caching the build process, so we can take advantage of the multi-stage build. We could use docker hub as a `registry source` to take advantage of the cache: https://docs.docker.com/reference/cli/docker/buildx/build/#cache-from

zhx828:
@jfkonecn @bprize15  sorry about the late reply. Even I could improve the build performance, loading docker image, and run docker containers still take more time then running the tests together though.

PR #416, opened 2024-08-15
adjust text area size anytime content size changes

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/506


src/main/webapp/app/shared/firebase/input/RealtimeBasicInput.tsx
```diff
@@ -115,6 +116,25 @@ const RealtimeBasicInput: React.FunctionComponent<IRealtimeBasicInput> = (props:
     };
   }, [firebasePath, db]);
 
+  useEffect(() => {
```
calvinlu3, line 119: Dependency array missing for use effect

calvinlu3 (approved):
One comment, otherwise LGTM!

src/main/webapp/app/shared/firebase/input/RealtimeBasicInput.tsx
bprize15: Thanks for catching this

PR #417, opened 2024-08-15
Do not show needs curation for categorical alterations

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/517

### DIFF
--- a/src/main/webapp/app/shared/badge/NotCuratableBadge.tsx
+++ b/src/main/webapp/app/shared/badge/NotCuratableBadge.tsx
@@ -5,6 +5,8 @@ import { componentInject } from '../util/typed-inject';
 import { getMutationName } from '../util/firebase/firebase-utils';
 import { RADIO_OPTION_NONE } from 'app/config/constants/constants';
 import { observer } from 'mobx-react';
+import { CategoricalAlterationType } from '../model/enumerations/categorical-alteration-type.model';
+import { parseAlterationName } from '../util/utils';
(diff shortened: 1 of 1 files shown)


src/main/webapp/app/shared/badge/NotCuratableBadge.tsx
```diff
@@ -22,6 +24,10 @@ const NotCuratableBadge: React.FunctionComponent<INotCuratableBadgeProps> = ({ m
     const mutationStrings: [string[], string[]] = [[], []];
     for (const name of mutationName.split(',')) {
       const trimmedName = name.trim();
+      if (Object.values(CategoricalAlterationType).includes(parseAlterationName(trimmedName)[0].alteration as CategoricalAlterationType)) {
```
calvinlu3, line 27: Currently, we for non-string mutations, only `Oncogenic Mutations` is not curatable. See util method: `isMutationEffectCuratable`. Maybe move the excludedMutations arrays out to a constants, so you can use it here?

calvinlu3 (changes requested):

src/main/webapp/app/shared/badge/NotCuratableBadge.tsx
bprize15: Ah gotcha, looking at the method, I could just use it directly

calvinlu3 (approved):
LGTM!

PR #420, opened 2024-08-15
Remove uuid from meta when newly created tumor has been deleted

calvinlu3:
fixes https://github.com/oncokb/oncokb-pipeline/issues/511

Was able to reproduce bug when:
Create a new tumor and delete it before review, the uuid is not cleared from meta collection.


zhx828 (approved):
LGTM, thanks. But for the past few prs, we haven't added any tests.

calvinlu3:
Good point, I will add some test cases specifically for checking if uuid has been removed from meta.

PR #421, opened 2024-08-15
Report all exceptions to setnry

zhx828:
The previous settings are copied over from oncokb public to avoid the noice. But we now want to see as much as possible. This fixes https://github.com/oncokb/oncokb-pipeline/issues/515

### DIFF
--- a/src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
+++ b/src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
@@ -18,10 +18,6 @@ import org.mskcc.oncokb.curation.config.application.ApplicationProperties;
 import org.mskcc.oncokb.curation.config.application.FrontendProperties;
 import org.mskcc.oncokb.curation.importer.CdxImporter;
 import org.mskcc.oncokb.curation.importer.Importer;
-import org.mskcc.oncokb.curation.service.InvalidPasswordException;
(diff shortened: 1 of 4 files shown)


src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
```diff
@@ -173,20 +162,15 @@ private static void initSentry(Environment env) {
...
-                options.addIgnoredExceptionForType(BadRequestAlertException.class);
-                options.addIgnoredExceptionForType(ResourceAccessException.class);
+                //                options.addIgnoredExceptionForType(IOException.class);
+                //                options.addIgnoredExceptionForType(HttpClientErrorException.class);
+                //                options.addIgnoredExceptionForType(HttpServerErrorException.class);
+                //                options.addIgnoredExceptionForType(ResponseStatusException.class);
+                //                options.addIgnoredExceptionForType(HttpRequestMethodNotSupportedException.class);
+                //                options.addIgnoredExceptionForType(InternalAuthenticationServiceException.class);
+                //                options.addIgnoredExceptionForType(BadRequestAlertException.class);
+                //                options.addIgnoredExceptionForType(ResourceAccessException.class);
```
zhx828, line 173: Let's keep these here for now. I will put them back based on the exceptions we receive in production.

calvinlu3 (approved):

PR #422, opened 2024-08-16
Parse comma seperated alterations and add info icon to add mutation modal input

calvinlu3:
fixes https://github.com/oncokb/oncokb-pipeline/issues/502

image


zhx828:
@calvinlu3 we can merge this first. But most importantly, we should include what alteration format is allowed in this input box, more details in SOP  Chapter 6: Table 3.1: OncoKB alteration nomenclature, style and formatting (page 162)

src/main/webapp/app/app.scss
```diff
@@ -429,3 +429,14 @@ a {
 .toastify-toast {
   display: inline-flex;
 }
+
+.scrollbar-wrapper {
```
zhx828, line 433: Does this have to be a global style?

zhx828 (approved):

src/main/webapp/app/app.scss
calvinlu3: Originally only had it in navigation sidebar scss, but figured that there could be other components that we might want to add this class.

PR #426, opened 2024-08-22
Add variant recommendation page

gebbing12:
Code for the Variant Recommendation Page - frontend and backend implementation


src/main/webapp/app/pages/variantRecommendation/FormatPercentage.tsx
```diff
@@ -0,0 +1,19 @@
+export function formatPercentage(value: number) {
+  if (value < 1e-4 && value > 0) {
+    let formatted = (value * 1000).toFixed(2);
+    formatted = formatted.replace(/\.?0+$/, '');
+    if (formatted.endsWith('.')) {
+      formatted = formatted.slice(0, -1);
+    }
```
jfkonecn, line 7: Is this to needed to remove trailing zeros? `let formatted = +(value * 1000).toFixed(2);` should give you the same result.

src/main/webapp/app/pages/variantRecommendation/FormatPercentage.tsx
```diff
@@ -0,0 +1,19 @@
+export function formatPercentage(value: number) {
+  if (value < 1e-4 && value > 0) {
+    let formatted = (value * 1000).toFixed(2);
```
jfkonecn, line 3: Why do you need to multiply by 1000?

src/main/webapp/app/pages/variantRecommendation/FormatPercentage.tsx
```diff
@@ -0,0 +1,19 @@
+export function formatPercentage(value: number) {
```
jfkonecn, line 1: Let's move this function to `src/main/webapp/app/shared/util/utils.tsx` since it doesn't really need to be in it's own file.

src/main/webapp/app/pages/variantRecommendation/FormatPercentage.tsx
```diff
@@ -0,0 +1,19 @@
...
+    let formatted = (value * 100).toFixed(2);
+    formatted = formatted.replace(/\.?0+$/, '');
+    if (formatted.endsWith('.')) {
+      formatted = formatted.slice(0, -1);
+    }
+    return `${formatted}%`;
+  }
+}
+
+export default formatPercentage;
```
jfkonecn, line 19: No need for a default export for this function.

src/main/webapp/app/pages/variantRecommendation/MutationAllFrequencyTab.tsx
```diff
@@ -0,0 +1,128 @@
...
+      filterMethod(filter, row) {
+        if (!filter.value) return true;
+        return String(row[filter.id]).toLowerCase().includes(filter.value.toLowerCase());
+      },
+    },
+    {
+      accessor: 'Number',
+      Header: 'Number',
+      Filter({ filter, onChange }) {
+        const filterValue = filter ? filter.value : ['', ''];
```
jfkonecn, line 72: You could also do something like this.
```ts
const [minValue, maxValue] = filter ? filter.value : ['', ''];
```

src/main/webapp/app/oncokb-commons/components/twoRangeSlider/TwoRangeSlider.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+interface RangeSliderProps {
+  min: number;
+  max: number;
+  range: [number, number];
+  step: number;
+  onChange: (newRange: [number, number]) => void;
+}
+
+export const TwoRangeSlider = ({ min, max, range, step, onChange }: RangeSliderProps) => {
```
jfkonecn, line 14: Why is this called `TwoRangeSlide`, why not just `Slider`?

src/main/webapp/app/pages/variantRecommendation/MutationDetailedFrequencyTab.tsx
```diff
@@ -0,0 +1,187 @@
...
+      Filter({ filter, onChange }) {
+        return (
+          <Input type="select" value={filter ? filter.value : ''} onChange={event => onChange(event.target.value)}>
+            {detailedCancerOptions}
+          </Input>
+        );
+      },
+      filterMethod(filter, row) {
+        if (!filter.value) return true;
+        return String(row[filter.id]).toLowerCase().includes(filter.value.toLowerCase());
```
jfkonecn, line 97: See if you can turn this into a function. I see the same repeated several times.

src/main/webapp/app/pages/variantRecommendation/MutationTypeFrequencyTab.tsx
```diff
@@ -0,0 +1,146 @@
...
+        if (!filter.value) return true;
+        return String(row[filter.id]).toLowerCase().includes(filter.value.toLowerCase());
+      },
+    },
+    {
+      accessor: 'Hugo_Symbol',
+      Header: 'Hugo Symbol',
+      Filter({ filter, onChange }) {
+        return <Input placeholder="Search" value={filter ? filter.value : ''} onChange={event => onChange(event.target.value)} />;
+      },
```
jfkonecn, line 64: Could this be made into a component? I see this code repeated in a few places.

src/main/webapp/app/pages/variantRecommendation/MutationTypeFrequencyTab.tsx
```diff
@@ -0,0 +1,146 @@
...
+      Header: 'Number',
+      Filter({ filter, onChange }) {
+        const filterValue = filter ? filter.value : ['', ''];
+        return (
+          <div style={{ display: 'flex', gap: '10px' }}>
+            <Input placeholder="Min" value={filterValue[0]} onChange={event => onChange([event.target.value, filterValue[1]])} />
+            <Input placeholder="Max" value={filterValue[1]} onChange={event => onChange([filterValue[0], event.target.value])} />
+          </div>
+        );
+      },
```
jfkonecn, line 103: Can this be made into a share component?

src/main/webapp/app/routes/routes.tsx
```diff
@@ -44,6 +45,7 @@ const Routes: React.FunctionComponent<IRoutesProps> = (props: IRoutesProps) => {
             <PrivateRoute exact path={PAGE_ROUTE.SEARCH} component={SearchPage} />
             <PrivateRoute path={PAGE_ROUTE.CURATION} component={CurationRoutes} hasAnyAuthorities={[AUTHORITIES.CURATOR]} />
             <PrivateRoute exact path={PAGE_ROUTE.ACCOUNT} component={Account} hasAnyAuthorities={[AUTHORITIES.USER]} />
+            <PrivateRoute exact path={PAGE_ROUTE.VARIANR_REC} component={VariantRecListPage} />
```
jfkonecn, line 48: Let's add `hasAnyAuthorities={[AUTHORITIES.CURATOR]}`.

Could you also add authorization to the backend too?

jfkonecn (changes requested):

src/main/webapp/app/oncokb-commons/components/twoRangeSlider/styles.module.scss
```diff
@@ -0,0 +1,107 @@
...
+  border: 0 none;
+}
+
+input[type='range']::-moz-range-thumb {
+  pointer-events: all;
+  width: var(--_thumb-size);
+  height: var(--_thumb-size);
+  border-radius: 0;
+  border: 0 none;
+}
```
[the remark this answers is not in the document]
gebbing12: This is to define the appearance of the slider thumb. -webkit-slider-thumb is for Chrome and Safari, while -moz-range-thumb is for Firefox.

src/main/webapp/app/pages/variantRecommendation/FormatPercentage.tsx
gebbing12: Because some numbers are very small, they can only be expressed in ‰ instead of %.

src/main/webapp/app/pages/variantRecommendation/FormatPercentage.tsx
gebbing12: I'd like to display 100% instead of 100.00%, as it looks better. However, if it doesn't matter, I can remove it.

src/main/webapp/app/pages/variantRecommendation/FormatPercentage.tsx
jfkonecn: Let's keep it as a percentage and not be per thousand. I think it'll confuse people if we change the units when they're reading through the rows.

src/main/java/org/mskcc/oncokb/curation/web/rest/VariantRecommendController.java
```diff
@@ -44,18 +45,17 @@ public ResponseEntity<List<JSONObject>> requestData(@PathVariable String filenam
...
                 JSONObject jsonObject = new JSONObject();
                 for (int i = 0; i < headers.length; i++) {
                     jsonObject.put(headers[i], "n/a".equals(data[i]) ? null : data[i]);
                 }
-                jsonList.add(jsonObject);
+                jsonArray.put(jsonObject);
             }
-
-            return ResponseEntity.ok(jsonList);
+            return ResponseEntity.ok(jsonArray.toString());
```
jfkonecn, line 58: Could you double that the content type is `Content-Type: application/json;` in the response header?

gebbing12: Done

PR #430, opened 2024-08-29
Use reviewed content only

jfkonecn:
I add a function, `useLastReviewedOnly`, to remove the unreviewed content.


zhx828 (changes requested):
Newly created section(new mutation for instance) without approval has been saved using `Save Gene` button.

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
```diff
@@ -1,8 +1,12 @@
 import _ from 'lodash';
 import { Drug, Gene, Mutation, Review, Treatment, Tumor, Vus, DrugCollection } from '../../model/firebase/firebase.model';
+import { useLastReviewedOnly } from '../core-submission-shared/core-submission-utils';
 
 export function getGeneData(geneData: Gene, onlyReviewedContent: boolean, drugList: DrugCollection): true | Gene {
-  const gene = _.cloneDeep(geneData);
+  const gene = onlyReviewedContent ? useLastReviewedOnly(geneData, true) : _.cloneDeep(geneData);
+  if (gene === undefined) {
+    return true;
```
zhx828, line 8: Why we returning boolean here?

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.spec.ts
```diff
@@ -123,6 +123,56 @@ const mockGene = createMockGene({
   }),
   mutations_uuid: '69e454db-db99-4979-a770-482f7937314b',
   mutations: [
+    createMockMutation({
```
zhx828, line 126: We should also add a test for new cancer type and new treatment.

src/main/webapp/app/shared/util/core-evidence-submission/core-evidence-submission.spec.ts
```diff
@@ -456,6 +459,7 @@ describe('getEvidence to submit to core', () => {
                     prognosticSummary_uuid: 'd9c991e4-1c13-4251-bd5a-fac0344b3470',
                     prognosticSummary: 'Prognostic Summary',
                     prognosticSummary_review: createMockReview({
+                      lastReviewed: 'Prognostic Summary Last Reviewed',
```
zhx828, line 462: What's the rational for testing Last Reviewed for only prognostic?

src/main/webapp/app/shared/util/core-submission-shared/core-submission-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export function useLastReviewedOnly<T>(obj: T, isSavingGene: boolean = false): T | undefined {
+  function useLastReviewedOnlyRec(curObj: T): T | undefined {
+    const newObj: T = {} as T;
+    for (const [key, value] of Object.entries(curObj as Record<string, unknown>)) {
+      if (key.endsWith('_review')) {
+        const temp: Review = Object.assign({} as Review, value);
+        if (!isSavingGene) {
+          delete temp.lastReviewed;
+        }
```
zhx828, line 11: Maybe this is copied over from the legacy, but do you know what this is for?

zhx828 (changes requested):

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
jfkonecn: because that's what the old code did for error cases, but I'll change that.

src/main/webapp/app/shared/util/core-evidence-submission/core-evidence-submission.spec.ts
jfkonecn: I felt like if I tried to test every case it'd get too complicate, but as far as why specifically prognostic summary? There isn't a real reason.

zhx828: Let's pick few relatively important ones? Could you do oncogenicity/Therapeutic Summary/Tx level

src/main/webapp/app/shared/util/core-submission-shared/core-submission-utils.ts
jfkonecn: It doesn't seem like this is needed. It'll remove it. I can't remember why I put it there in the first place.

src/main/webapp/app/shared/util/core-drive-annotation-submission/core-drive-annotation-submission.ts
```diff
@@ -93,12 +89,8 @@ export function getVUSData(vus: Vus[]) {
...
 }
 
-function shouldExclude(onlyReviewedContent: boolean, reviewObj: Review | undefined) {
-  return (
-    reviewObj &&
-    ((onlyReviewedContent && reviewObj.added === true && reviewObj.promotedToMutation === true && reviewObj.initialUpdate === true) ||
-      (!onlyReviewedContent && reviewObj.removed === true))
-  );
+function shouldExclude(reviewObj: Review | undefined) {
+  return reviewObj && reviewObj.added === true && reviewObj.promotedToMutation === true && reviewObj.initialUpdate === true;
```
zhx828, line 93: Based on the comment in firebase.model.ts, `added` and `promotedToMutation` can coexist. And look at firebase-gene-service, when promotedToMutation is true, added is always true. Could you take a look?

jfkonecn: Yes this appears to be the case. I also couldn't find where `intialUpdate` is ever `true`. Does that mean this should exclude logic will always be `false`?

zhx828: `return reviewObj && (reviewObj.added === true || reviewObj.promotedToMutation === true || reviewObj.initialUpdate === true);` based on the conversation, we should update to this.
Could you also update the comment to reflect what Calvin said?

zhx828 (approved):
Thank you

PR #434, opened 2024-09-09
Avoid fetching management info repeatedly

zhx828:
- [Avoid fetching management info repeatedly](https://github.com/oncokb/oncokb-transcript/commit/7f64f9b295d2cdc9f8ad9b736f58eccc0c87bfcc)

- [Do not rerender the side bar when loading session](https://github.com/oncokb/oncokb-transcript/commit/6575c8c67c3cc72b63cad11e62478c989091065b) 
    The getSession is an async method which updates the loading status.

### DIFF
--- a/src/main/webapp/app/app.tsx
+++ b/src/main/webapp/app/app.tsx
@@ -42,9 +42,8 @@ const App: React.FunctionComponent<IAppProps> = (props: IAppProps) => {
             pauseOnFocusLoss
           />
           <BetaSiteMessage />
-          {props.loadingAuth ? (
-            <LoadingIndicator isLoading size={LoaderSize.LARGE} center={true} />
-          ) : (
+          {props.loadingAuth && <LoadingIndicator isLoading size={LoaderSize.LARGE} center={true} />}
+          {props.authAccount && (
             <div>
               {props.isAuthorized && <NavigationSidebar />}
               <div className="app-center-content-wrapper" style={{ margin: props.centerContentMargin }}>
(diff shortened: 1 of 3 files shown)


src/main/webapp/app/app.tsx
```diff
@@ -42,9 +42,8 @@ const App: React.FunctionComponent<IAppProps> = (props: IAppProps) => {
             pauseOnFocusLoss
           />
           <BetaSiteMessage />
-          {props.loadingAuth ? (
-            <LoadingIndicator isLoading size={LoaderSize.LARGE} center={true} />
-          ) : (
+          {props.loadingAuth && <LoadingIndicator isLoading size={LoaderSize.LARGE} center={true} />}
```
zhx828, line 45: @calvinlu3 I think this solution is a bit hacky, let me know how you think.

src/main/webapp/app/app.tsx
```diff
@@ -63,6 +62,7 @@ const mapStoreToProps = ({ authStore, layoutStore, firebaseAppStore }: IRootStor
   authorities: authStore.account.authorities,
   isCurator: hasAnyAuthority(authStore.account.authorities ?? [], [AUTHORITIES.CURATOR]),
   loadingAuth: authStore.loading,
+  authAccount: authStore.account,
```
calvinlu3, line 65: Use `authStore.isAuthenticated` instead?

src/main/webapp/app/app.tsx
calvinlu3: This looks fine to me. What issue did rendering the AppRoutes cause?

src/main/webapp/app/app.tsx
zhx828: @calvinlu3 I think it should be ok, but based on the logic in line 48, we still want to render the routes even the user is not authorized?

calvinlu3: Yeah there are some routes that should be available when unauthorized like login and logout. The other routes are wrapped as PrivateRoutes, so they will check for authorities.

calvinlu3 (approved):

src/main/webapp/app/app.tsx
zhx828: If that's the case, I can't use isAuthenticated then. I will not render the login/logout.

PR #439, opened 2024-09-17
Feature/fix submit to core

jfkonecn:
@zhx828 it looks like nothing is submitted when an accept all is done. Is that behavior we want?


src/main/webapp/app/pages/curation/review/ReviewPage.tsx
```diff
@@ -125,7 +121,7 @@ const ReviewPage: React.FunctionComponent<IReviewPageProps> = (props: IReviewPag
         isGermline,
         isAcceptAll: true,
         gene: geneData,
-        entrezGeneId: props.geneEntity.entrezGeneId,
+        entrezGeneId: geneEntity?.entrezGeneId as number,
```
jfkonecn, line 124: No longer using the prop entity since it's already fetched within this component.

src/main/webapp/app/shared/util/core-submission-shared/core-submission-utils.ts
```diff
@@ -1,33 +1,42 @@
...
+  return valuePath !== undefined && path.length > 0 && (path.startsWith(valuePath) || valuePath.startsWith(path));
+}
+
+export function useLastReviewedOnly<T>(obj: T, valuePath?: string): T | undefined {
+  function useLastReviewedOnlyRec(curObj: T, parentPath: string): T | undefined {
     const newObj: T = {} as T;
     for (const [key, value] of Object.entries(curObj as Record<string, unknown>)) {
+      const currentPath = `${parentPath}${parentPath.length > 0 ? '/' : ''}${key}`;
+      const protectCurrent = shouldProtect(valuePath, currentPath);
+      const protectParent = shouldProtect(valuePath, parentPath);
```
jfkonecn, line 13: Need to protect the whole object if the parent is in the valuePath since we don't want to drop the whole object.

src/main/webapp/app/shared/util/core-submission-shared/core-submission-utils.ts
```diff
@@ -1,33 +1,42 @@
...
           return undefined;
         }
-        const arr: T[] = [];
+        const arr: (T | undefined)[] = [];
+        let i = 0;
         for (const arrObj of value) {
-          const temp = useLastReviewedOnlyRec(arrObj);
-          if (temp !== undefined) {
+          const temp = useLastReviewedOnlyRec(arrObj, `${currentPath}/${i}`);
+          if (temp !== undefined || protectParent) {
```
jfkonecn, line 27: keep undefined value in the array so that the index still exists for a path like `mutation/11/....`

src/main/webapp/app/shared/util/core-submission-shared/core-submission-utils.ts
```diff
@@ -1,40 +1,121 @@
...
+    if (shouldProtect) {
+      return true;
+    }
+  }
+  return false;
+}
+
+/**
+ * Recursively filters the provided object to include only the last reviewed elements
+ * or elements that where just accepted.
```
zhx828, line 26:
```suggestion
 * or elements that were just accepted.
```

zhx828 (approved):
Looking good!

PR #441, opened 2024-09-18
Parse protein change case insensitive

zhx828:
This is related to https://github.com/oncokb/oncokb/issues/3760


src/main/java/org/mskcc/oncokb/curation/util/parser/ProteinChangeParser.java
```diff
@@ -0,0 +1,360 @@
...
+            termsToCheck.put("fs", FEATURE_TRUNCATION);
+            termsToCheck.put("trunc", FEATURE_TRUNCATION);
+            termsToCheck.put("dup", INFRAME_INSERTION);
+            termsToCheck.put("mut", ANY);
+
+            MutationConsequence mutationConsequence = termsToCheck.get(variant);
+            if (mutationConsequence != null) {
+                revisedProteinChange += m.group(1).toUpperCase();
+                revisedProteinChange += m.group(2) + "_";
+                if (m.group(3) != null) {
```
calvinlu3, line 256: Since group 3 is `([A-Z]?)`, I believe it will return empty string if not present instead of null. If it was written as `([A-Z])?` then group 3 can be null.

Can remove this null check.

src/test/resources/data/test_parse_protein_change.csv
```diff
@@ -0,0 +1,92 @@
...
+A149*,stop_gained,A,*,149,149
+149*,stop_gained,,*,149,149
+*149*,stop_lost,*,*,149,149
+
+# extension
+*757Kext*36,stop_lost,*,,757,757
+757Kext*36,stop_lost,*,,757,757
+*757ext*36,stop_lost,*,,757,757
+*757Kext*,stop_lost,*,,757,757
+*757Kext*?,stop_lost,*,,757,757
```
calvinlu3, line 76: Do you have some examples for `M?1ext(-[0-9]+)?` ?

zhx828: Good catch. We don't. I'm digging more details on this one, will update.

calvinlu3 (approved):

PR #442, opened 2024-09-19
break in middle of word to fix collapsible title overlfow

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/541


calvinlu3 (approved):
Thanks!

PR #443, opened 2024-09-20
Add mutations sorting by firebase index

bprize15:


src/main/webapp/app/pages/curation/header/MutationsSectionHeader.tsx
```diff
@@ -46,6 +46,7 @@ export enum SortOptions {
   LAST_MODIFIED = 'Last Modified',
   POSITION_INCREASING = 'Position Increasing',
   POSITION_DECREASING = 'Position Decreasing',
+  FIREBASE = 'Firebase',
```
zhx828, line 49: I assume this will make the sort method returning 0 and it will use the order in the list?

bprize15: Yeah exactly

zhx828 (approved):
Thank you!

PR #444, opened 2024-09-20
Delete mutation/tumor/treatment as last step to avoid stale index

calvinlu3:
fixes https://github.com/oncokb/oncokb-pipeline/issues/551

### DIFF
--- a/src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
+++ b/src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
@@ -159,6 +159,7 @@ export class FirebaseGeneReviewService {
     }
 
     // We are deleting last because the indices will change after deleting from array.
+    let hasDeletion = false;
     try {
       // Todo: We should use multi-location updates for deletions once all our arrays use firebase auto-generated keys
       // instead of using sequential number indices.
@@ -168,8 +169,13 @@ export class FirebaseGeneReviewService {
         FIREBASE_LIST_PATH_TYPE.MUTATION_LIST,
       ]) {
         for (const [firebasePath, deleteIndices] of Object.entries(itemsToDelete[pathType])) {
+          hasDeletion = true;
           await this.firebaseRepository.deleteFromArray(firebasePath, deleteIndices);
         }
+      }
+      // If user accepts a deletion individually, we need to refresh the ReviewPage with the latest data to make sure the indices are up to date.
+      if (reviewLevels.length === 1 && hasDeletion) {
+        return { shouldRefresh: true };
       }
     } catch (error) {
       throw new SentryError('Failed to accept deletions in review mode', { hugoSymbol, reviewLevels, isGermline, itemsToDelete });
--- a/src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
+++ b/src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
@@ -6,7 +6,12 @@ import { FirebaseRepository } from 'app/stores/firebase/firebase-repository';
 import _ from 'lodash';
 import { Review } from '../../shared/model/firebase/firebase.model';
 import { buildHistoryFromReviews } from '../../shared/util/firebase/firebase-history-utils';
-import { extractArrayPath, parseFirebaseGenePath } from '../../shared/util/firebase/firebase-path-utils';
+import {
+  extractArrayPath,
+  FIREBASE_PATH_TYPE,
+  getFirebasePathType,
+  parseFirebaseGenePath,
+} from '../../shared/util/firebase/firebase-path-utils';
 import {
   ReviewLevel,
   TumorReviewLevel,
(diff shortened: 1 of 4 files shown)


src/main/webapp/app/shared/util/firebase/firebase-path-utils.ts
```diff
@@ -42,3 +42,20 @@ export const extractArrayPath = (valuePath: string) => {
   const firebaseArrayPath = pathParts.join('/');
   return { firebaseArrayPath, deleteIndex };
 };
+
+export enum FIREBASE_PATH_TYPE {
```
zhx828, line 46: Name is too generic. Maybe 
```suggestion
export enum FIREBASE_LIST_PATH_TYPE {
```

src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
```diff
@@ -115,17 +126,15 @@ export class FirebaseGeneReviewService {
       } else if (isDeleteReview(reviewLevel)) {
         const { firebaseArrayPath, deleteIndex } = extractArrayPath(reviewLevel.valuePath);
         const firebasePath = geneFirebasePath + '/' + firebaseArrayPath;
+        const firebasePathType = getFirebasePathType(firebaseArrayPath + '/' + deleteIndex);
+        if (firebasePathType !== undefined) {
+          const innerMap = itemsToDelete[firebasePathType];
```
zhx828, line 131: Is `innerMap` used?

src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
```diff
@@ -140,7 +149,28 @@ export class FirebaseGeneReviewService {
...
+    try {
+      // Todo: We should use multi-location updates for deletions once all our arrays use firebase auto-generated keys
+      // instead of using sequential number indices.
+      /* eslint-disable no-console */
+      console.log(itemsToDelete);
+      for (const pathType of [FIREBASE_PATH_TYPE.TREATMENT_LIST, FIREBASE_PATH_TYPE.TUMOR_LIST, FIREBASE_PATH_TYPE.MUTATION_LIST]) {
+        for (const [firebasePath, deleteIndices] of Object.entries(itemsToDelete[pathType])) {
+          await this.firebaseRepository.deleteFromArray(firebasePath, deleteIndices);
+        }
+      }
```
zhx828, line 171: I see here we are trying to delete deeper list first to avoid mismatch. But how are you addressing deleting/updating the next index?

zhx828 (changes requested):

zhx828:
We need some comments about these logics. We ned test to make sure after approving deletion will not affect the update on other indices.

src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
calvinlu3: Good question. 
`firebaseRepository.deleteFromArray(firebasePath, deleteIndices)` will perform the deletion of the array of index in a transaction. The index will never be stale because we loop through the original list to construct a new firebase list. 

Check out the code: https://github.com/oncokb/oncokb-transcript/blob/rc/src/main/webapp/app/stores/firebase/firebase-repository.ts#L83-L92

src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
calvinlu3: I use it on the next line, so I don't have to do `itemsToDelete[firebasePathType][firebasepath]`

src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
```diff
@@ -140,7 +149,28 @@ export class FirebaseGeneReviewService {
...
+        updateObject,
+      });
+    }
+
+    // We are deleting last because the indices will change after deleting from array.
+    try {
+      // Todo: We should use multi-location updates for deletions once all our arrays use firebase auto-generated keys
+      // instead of using sequential number indices.
+      /* eslint-disable no-console */
+      console.log(itemsToDelete);
```
bprize15, line 166: Remove

bprize15 (approved):
One minor comment, otherwise LGTM

zhx828 (approved):

src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
```diff
@@ -168,9 +169,14 @@ export class FirebaseGeneReviewService {
         FIREBASE_LIST_PATH_TYPE.MUTATION_LIST,
       ]) {
         for (const [firebasePath, deleteIndices] of Object.entries(itemsToDelete[pathType])) {
+          hasDeletion = true;
           await this.firebaseRepository.deleteFromArray(firebasePath, deleteIndices);
         }
       }
+      // If user accepts a deletion individually, we need to refresh the ReviewPage with the latest data to make sure the indices are up to date.
+      if (reviewLevels.length === 1 && hasDeletion) {
```
bprize15, line 177: Why do we only refresh if the length is 1? If a review level corresponds to some change that needs to be reviewed, wouldn't we want to refresh if length > 0?

calvinlu3: The only time reviewLevels length is > 1 is when we use the accept all feature. In that case a refresh is not needed because everything should be accepted all together.

zhx828 (approved):
Once Ben's comment is addressed, it's good to go.

bprize15 (approved):
LGTM

PR #447, opened 2024-09-25
Fixed data validation tool

jfkonecn:
Looks like the websocket request was being redirected to the index.html file.
I'm not sure why the toString needed to be changed to getPath. 
Maybe something changed in the spring boot version?


zhx828 (approved):
Thanks!

PR #454, opened 2024-10-03
Make core API call non-blocking

jfkonecn:
closes oncokb/oncokb-pipeline#569

I added the feature flag so I wouldn't miss a spot when I revert this change.

closes oncokb/oncokb-pipeline#570

I added a loading indicator as well.


src/main/webapp/app/pages/curation/review/ReviewPage.tsx
```diff
@@ -206,6 +206,7 @@ const ReviewPage: React.FunctionComponent<IReviewPageProps> = (props: IReviewPag
           </Row>
         </>
       )}
+      {(isAcceptingAll || isAccepting) && <LoadingIndicator key={'curation-page-loading'} size={LoaderSize.LARGE} center isLoading />}
```
zhx828, line 209: This is new to me. The loader is on top of the review page. What's your thought on this?

src/main/webapp/app/shared/feature-flags.ts
```diff
@@ -0,0 +1 @@
+export const STOP_REVIEW_IF_CORE_SUBMISSION_FAILS: boolean = false;
```
zhx828, line 1: Should this be in appConfig.ts?

src/main/webapp/app/pages/curation/review/ReviewPage.tsx
jfkonecn: Yes it's at the top. I have no real preference on this. You have a better eye for UI than I do.

src/main/webapp/app/shared/feature-flags.ts
jfkonecn: I think you could make an argument either way for this. If you feel like you'll want to switch the behavior without creating a new build, then yes. If you feel like we'll make a code change first then rolls this out, then it doesn't make a different. This is entirely for myself to track the places I need to rollback.

src/main/webapp/app/pages/curation/review/ReviewPage.tsx
zhx828: We could hide all review content when accepting all, and show a loader on the white background. Could you try that? Or without.

src/main/webapp/app/shared/feature-flags.ts
zhx828: Let's move this to AppConfig to consolidate.

jfkonecn:
@zhx828 I addressed you comments.
I also decided to fix the merge conflicts here as well since this is probably going to be the last PR before this goes into RC.

https://github.com/oncokb/oncokb-transcript/pull/454/commits/298f9c8e7ee239a96ab4880676243226d5369424
https://github.com/oncokb/oncokb-transcript/pull/454/commits/5ebc9a8770b2bcbe5b3e3690e855e3ec50081c91

PR #461, opened 2024-10-22
Update CDx curation panel to allow combination drugs

calvinlu3:
As a quick fix, we will convert the drug select on CDx panel to input combination drugs instead of list of treaments (comma seperated). 
There will be a followup PR to address adding both combination drugs and multiple treatments. (Same interface as add therapy modal on curation page) https://github.com/oncokb/oncokb-pipeline/issues/588


zhx828 (approved):
LGTM

PR #462, opened 2024-10-24
Add reject to created review

calvinlu3:
Fixes https://github.com/oncokb/oncokb-pipeline/issues/581

### Changes
- Added reject button to `Created` or `Promoted` review collapsibles
image

- On reject promoted mutation, add variants back to VUS list


src/main/webapp/app/pages/curation/collapsible/ReviewCollapsible.tsx
```diff
@@ -239,21 +239,29 @@ export const ReviewCollapsible = ({
...
+        onClick={() => {
+          handleActionClick(ActionType.REJECT);
+        }}
+        disabled={disableActions}
+      />
+    );
     if (isCreateReview(rootReview)) {
-      // Reviewable content should be reviewed individually under a create event
-      return undefined;
+      if (rootReview.nestedUnderCreateOrDelete) {
```
bprize15, line 253: Just wondering why we don't allow deleting a nested creation. For example, when I create a mutation and then a cancer type under it, why do we not allow rejecting the entire cancer type?

src/main/webapp/app/pages/curation/review/ReviewPage.tsx
```diff
@@ -50,6 +50,7 @@ const ReviewPage: React.FunctionComponent<IReviewPageProps> = (props: IReviewPag
   const [isAcceptingAll, setIsAcceptingAll] = useState(false);
 
   const [isAccepting, setIsAccepting] = useState(false);
+  const [isRejecting, setIsRejecting] = useState(false);
```
bprize15, line 53: Could we combine these two? It seems we just want to disable actions when this is happening?

Could we do something similar with isAcceptingAll? I see you hide the page when this is happening, but is it enough to just disable the buttons. If this is the case, we could remove these 3 state variables and combine them into a `loading` state

src/main/webapp/app/pages/curation/collapsible/ReviewCollapsible.tsx
calvinlu3: Yes, I think nested creations should be deletable as well. However, let's consider the scenario you described. If we reject the cancer type creation (which is the only nested review under the mutation), then the mutation gets automatically accepted. This is due to the behavior we implemented that a created entity will be auto-accepted when all items have been reviewed. 

We have a ticket to add a `isReviewPending` status to review object, so that we can keep track of the items that are pending acceptance.

src/main/webapp/app/pages/curation/review/ReviewPage.tsx
calvinlu3: I can combine the isAccepting and isRejecting into one status. However, we do need isAcceptingAll because one of the FirebaseGeneReviewService method needs to know that status.

src/main/webapp/app/pages/curation/collapsible/ReviewCollapsible.tsx
bprize15: That's a good point. Thanks for letting me know.

bprize15 (approved):
LGTM, thanks!

PR #463, opened 2024-10-25
Save user image to database and display if available

calvinlu3:
### Changes
- Change Keycloak Google IDp `Sync Mode` setting from `import` -> `force`. This will send a `picture` attribute as part of OAuth token, so that we can save to our database.
- Display user profile image if available


zhx828 (approved):
Thanks!

PR #466, opened 2024-10-30
Feature Flag

gebbing12:
Fixes https://github.com/oncokb/oncokb-pipeline/issues/585
Add Java, changelog files for feature flag

### DIFF
--- a/src/main/resources/config/liquibase/changelog/20241024231608_added_entity_FeatureFlag.xml
+++ b/src/main/resources/config/liquibase/changelog/20241024231608_added_entity_FeatureFlag.xml
@@ -1,0 +1,45 @@
+<?xml version="1.0" encoding="utf-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd
+                        http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+
+    <!--
+        Added the entity FeatureFlag.
+    -->
+    <changeSet id="20241024231608-1" author="jhipster">
+        <createTable tableName="feature_flag">
+            <column name="id" type="bigint" autoIncrement="true">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="name" type="varchar(50)">
+                <constraints nullable="false" />
+            </column>
+            <column name="description" type="varchar(200)">
+                <constraints nullable="true" />
+            </column>
+            <column name="enabled" type="boolean">
+                <constraints nullable="true" />
+            </column>
+            <!-- jhipster-needle-liquibase-add-column - JHipster will add columns here -->
+        </createTable>
+    </changeSet>
+
+    <changeSet id="20241024231608-1-relations" author="jhipster">
+
+        <createTable tableName="rel_feature_flag__user">
+            <column name="user_id" type="bigint">
+                <constraints nullable="false"/>
+            </column>
+            <column name="feature_flag_id" type="bigint">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+
+        <addPrimaryKey columnNames="feature_flag_id, user_id" tableName="rel_feature_flag__user"/>
+    </changeSet>
+
+    <!-- jhipster-needle-liquibase-add-changeset - JHipster will add changesets here -->
+</databaseChangeLog>
--- a/src/main/resources/config/liquibase/changelog/20241024231608_added_entity_FeatureFlag.xml
+++ b/src/main/resources/config/liquibase/changelog/20241024231608_added_entity_FeatureFlag.xml
@@ -9,19 +9,19 @@
     <!--
         Added the entity FeatureFlag.
     -->
-    <changeSet id="20241024231608-1" author="jhipster">
+    <changeSet id="20241024231608-1" author="oncokb">
         <createTable tableName="feature_flag">
             <column name="id" type="bigint" autoIncrement="true">
                 <constraints primaryKey="true" nullable="false"/>
             </column>
-            <column name="name" type="varchar(50)">
+            <column name="name" type="varchar(255)">
                 <constraints nullable="false" />
             </column>
-            <column name="description" type="varchar(200)">
+            <column name="description" type="varchar(255)">
                 <constraints nullable="true" />
             </column>
             <column name="enabled" type="boolean">
-                <constraints nullable="true" />
+                <constraints nullable="false" default="false"/>
             </column>
             <!-- jhipster-needle-liquibase-add-column - JHipster will add columns here -->
         </createTable>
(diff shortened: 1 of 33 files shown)


src/main/resources/config/liquibase/changelog/20241024231608_added_entity_FeatureFlag.xml
```diff
@@ -0,0 +1,45 @@
...
+    -->
+    <changeSet id="20241024231608-1" author="jhipster">
+        <createTable tableName="feature_flag">
+            <column name="id" type="bigint" autoIncrement="true">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="name" type="varchar(50)">
+                <constraints nullable="false" />
+            </column>
+            <column name="description" type="varchar(200)">
```
zhx828, line 20: Our default is varchar(255)

src/main/resources/config/liquibase/changelog/20241024231608_added_entity_FeatureFlag.xml
```diff
@@ -0,0 +1,45 @@
...
+                <constraints nullable="true" />
+            </column>
+            <column name="enabled" type="boolean">
```
zhx828, line 23: I think this should not be null and should be false by default.

zhx828 (changes requested):
Are most of the files auto generated or copied over?

gebbing12:
> Are most of the files auto generated or copied over?

Yes, all Java and XML files are generated by JHipster. However, I modified the added_entity_FeatureFlag.xml file to set the composite primary key for rel_feature_flag__user and adjusted the length requirements for name and description in feature_flag. I also commented getAllFeatureFlagsByUserIsEqualToSomething() function in FeatureFlagResourceIT.java since it uses UserResourceIT class, which is not included in the project.

src/test/java/org/mskcc/oncokb/curation/web/rest/FeatureFlagResourceIT.java
```diff
@@ -0,0 +1,766 @@
...
+    //     featureFlag.addUser(user);
+    //     featureFlagRepository.saveAndFlush(featureFlag);
+    //     Long userId = user.getId();
+
+    //     // Get all the featureFlagList where user equals to userId
+    //     defaultFeatureFlagShouldBeFound("userId.equals=" + userId);
+
+    //     // Get all the featureFlagList where user equals to "invalid-id"
+    //     defaultFeatureFlagShouldNotBeFound("userId.equals=" + "invalid-id");
+    // }
```
jfkonecn, line 478: Was it not possible to get this test to pass?

gebbing12: The main problem is that we do not have the UserResourceIT class, since we do not need to use this function now, I have just committed it.

jfkonecn (approved):

src/main/webapp/app/modules/administration/user-management/user-management-update.tsx
```diff
@@ -45,16 +51,23 @@ export const UserManagementUpdate = (props: IUserManagementUpdateProps) => {
...
           ...user,
           authorities: user?.authorities?.map(auth => ({ label: auth, value: auth })),
+          featureFlags: user?.featureFlags?.map(f => ({ label: f.name, value: f })),
         };
-  };
+  }, [user]);
+
+  useEffect(() => {
+    setIsDev(props.user?.authorities?.includes(USER_AUTHORITY.ROLE_DEV));
+  }, [user]);
```
calvinlu3, line 68: Let's destructure the `props` on line 16, so you don't need to call `props.[variable]` everytime. Can get rid of line 54 after making that change.

src/main/webapp/app/modules/administration/user-management/user-management-update.tsx
```diff
@@ -124,10 +146,12 @@ const mapStoreToProps = (storeState: IRootStore) => ({
   user: storeState.userStore.entity,
   loading: storeState.userStore.loading,
   updating: storeState.userStore.updating,
+  featureFlags: storeState.featureFlagStore.entities,
   getUser: storeState.userStore.getEntity,
   updateUser: storeState.userStore.updateEntity,
   createUser: storeState.userStore.createEntity,
   reset: storeState.userStore.reset,
+  getFeatureFlags: storeState.featureFlagStore.getEntities,
```
calvinlu3, line 162: Add a prop here: `isDev: hasAnyAuthority(authStore.account.authorities, [AUTHORITIES.DEV])` to determine if user is a dev instead of the isDev state variable

src/main/webapp/app/modules/administration/user-management/user-management-update.tsx
```diff
@@ -110,7 +123,16 @@ export const UserManagementUpdate = (props: IUserManagementUpdateProps) => {
...
+                onChange={(selectedOption, actionMeta) => {
+                  if (actionMeta.action === 'remove-value' && actionMeta.removedValue?.value === USER_AUTHORITY.ROLE_DEV) {
+                    setIsDev(false);
+                  }
+                  if (actionMeta.action === 'select-option' && selectedOption.some(option => option.value === USER_AUTHORITY.ROLE_DEV)) {
+                    setIsDev(true);
+                  }
+                }}
               />
+              {isDev && <ValidatedSelect isMulti label="Feature Flag" name={'featureFlags'} options={featureFlagOptions} />}
```
calvinlu3: Instead of isDev state variable, I would just check if the user has ROLE_DEV authority. See my comment on line 154.

src/main/webapp/app/modules/administration/user-management/user-management-update.tsx
```diff
@@ -110,7 +123,16 @@ export const UserManagementUpdate = (props: IUserManagementUpdateProps) => {
...
                 options={authorityOptions}
                 defaultValue={[{ label: USER_AUTHORITY.ROLE_USER, value: USER_AUTHORITY.ROLE_USER }]}
+                onChange={(selectedOption, actionMeta) => {
+                  if (actionMeta.action === 'remove-value' && actionMeta.removedValue?.value === USER_AUTHORITY.ROLE_DEV) {
+                    setIsDev(false);
+                  }
+                  if (actionMeta.action === 'select-option' && selectedOption.some(option => option.value === USER_AUTHORITY.ROLE_DEV)) {
+                    setIsDev(true);
+                  }
+                }}
```
calvinlu3, line 133: I see that you are trying to show the Feature Flag dropdown only if the user is a dev. 
Note that the user management page is only accessible to admin (ROLE_ADMIN).
This means there is two scenarios:
1. An admin+dev user is editing another user (who has ROLE_DEV). If admin removes DEV role, it is still fine to show the feature flag dropdown because whether you show/not show the dropdown depends on if the current user is a dev or not.
2. An admin+dev user is editing their own account. They remove their own DEV role, but once you click the `Save` button, the page will refresh and the feature flag dropdown will not be visible anymore. In other words, the user is still a dev user until they save the DEV role removal changes.

tldr; I think it is fine to not have this onChange handler

calvinlu3: Looping in @jfkonecn to see what you think because I realized that this code was already approved.

calvinlu3 (changes requested):
Some small comments. The new changes do look good. @jfkonecn We should test this branch on beta.curation.org.

src/main/webapp/app/modules/administration/user-management/user-management-update.tsx
jfkonecn: I think it's fine to leave it as is. If it ends up being a problem it's easy to revert.

calvinlu3 (approved):
LGTM

PR #468, opened 2024-11-04
Give option to create gene when importing data

zhx828:
We current indicate the gene does not exist when importing.
This fixes https://github.com/oncokb/oncokb-pipeline/issues/599

### DIFF
--- a/src/main/webapp/app/components/tabs/curation-data-import-tab/import-methods.tsx
+++ b/src/main/webapp/app/components/tabs/curation-data-import-tab/import-methods.tsx
@@ -97,6 +97,7 @@ const pathogenicityCheck = (pathogenicity: string): DataImportStatus => {
 export const geneCheck = async (
   firebaseGeneService: FirebaseGeneService,
   isGermline: boolean,
+  createGene: boolean,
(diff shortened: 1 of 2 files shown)


src/main/webapp/app/components/tabs/curation-data-import-tab/import-methods.tsx
```diff
@@ -108,8 +109,13 @@ export const geneCheck = async (
...
     } else {
-      status.status = 'error';
-      status.message = 'Gene does not exist. Please create before importing';
+      if (createGene) {
+        await firebaseGeneService.createGene(hugoSymbol, isGermline);
+        return onGeneExists();
+      } else {
+        status.status = 'error';
+        status.message = 'Gene does not exist. Please create before importing';
+      }
```
calvinlu3, line 118: Is there a reason to not auto create gene?

zhx828: Could be typo, could be other reasons.

calvinlu3 (approved):
LGTM

PR #469, opened 2024-11-07
Create gene on gene link click if does not exist

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/604

### DIFF
--- a/src/main/webapp/app/pages/curation/GeneListPage.tsx
+++ b/src/main/webapp/app/pages/curation/GeneListPage.tsx
@@ -77,7 +77,7 @@ const GeneListPage = (props: IGeneListPage) => {
               event.preventDefault();
               if (props.firebaseDb) {
                 try {
-                  await createGeneIfDoesNotExist(cell.value, !isGermline, props.firebaseDb, props.createGene);
+                  await createGeneIfDoesNotExist(cell.value, isGermline, props.firebaseDb, props.createGene);
                   window.location.href = link;
                 } catch (error) {
                   notifyError(error);
--- a/src/main/webapp/app/pages/curation/GeneListPage.tsx
+++ b/src/main/webapp/app/pages/curation/GeneListPage.tsx
@@ -4,7 +4,7 @@ import { IRootStore } from 'app/stores';
 import { Col, Row } from 'reactstrap';
 import LoadingIndicator, { LoaderSize } from 'app/oncokb-commons/components/loadingIndicator/LoadingIndicator';
 import { createGeneIfDoesNotExist, geneNeedsReview } from 'app/shared/util/firebase/firebase-utils';
-import { Link, RouteComponentProps, generatePath } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


src/main/webapp/app/shared/util/firebase/firebase-utils.tsx
```diff
@@ -961,3 +963,16 @@ export function areCancerTypePropertiesEqual(a: string | undefined, b: string |
 export function isStringEmpty(string: string | undefined | null) {
   return string === '' || _.isNil(string);
 }
+
+export async function createGeneIfDoesNotExist(
+  hugoSymbol: string,
+  isSomatic: boolean,
```
zhx828, line 969: I think most of our methods use isGermline, could you update?

zhx828 (changes requested):
One minor request

src/main/webapp/app/pages/curation/GeneListPage.tsx
```diff
@@ -67,7 +68,26 @@ const GeneListPage = (props: IGeneListPage) => {
       accessor: 'hugoSymbol',
       Header: 'Hugo Symbol',
       Cell(cell: { value: string }): any {
-        return <Link to={getCurationPageLink(cell.value, isGermline)}>{cell.value}</Link>;
+        const link = getCurationPageLink(cell.value, isGermline);
+
+        return (
+          <Link
+            to={undefined}
```
calvinlu3, line 75: What do you think about converting this to a Button that is styled like a link?

src/main/webapp/app/pages/curation/GeneListPage.tsx
```diff
@@ -67,7 +68,26 @@ const GeneListPage = (props: IGeneListPage) => {
...
+            onClick={async event => {
+              event.preventDefault();
+              if (props.firebaseDb) {
+                try {
+                  await createGeneIfDoesNotExist(cell.value, isGermline, props.firebaseDb, props.createGene);
+                  window.location.href = link;
```
calvinlu3, line 81: Do you think using `useHistory` and `history.push()` would be better? It should be faster because it is not doing a full page reload. For instance, I believe the navigation sidebar should be rendered pretty quickly. Worth a shot to see if a there's a difference.

src/main/webapp/app/pages/curation/GeneListPage.tsx
bprize15: I prefer to keep it as a link. That way any styling we make to the link element also apply here.

src/main/webapp/app/pages/curation/GeneListPage.tsx
bprize15: Good point, definitely feels better that way.

calvinlu3 (approved):
LGTM!

zhx828 (approved):
Thank you

PR #470, opened 2024-11-08
Gene header redesign

calvinlu3:
Fixes https://github.com/oncokb/oncokb-pipeline/issues/592

## Changes
- Database `Flag Table` change: Split ONCOKB flag in ONCOKB_SOMATIC and ONCOKB_GERMLINE
- Redesign gene header
   - Some genes are still missing some transcripts and refseq information that will be backfilled in https://github.com/oncokb/oncokb-pipeline/issues/262
- Added GeneticTypeTabs
- Include release and review status
- Include current reviewer and last editor
- Move `isGermline` to router store, so we don't need to pass as prop to each component.

### When another person is reviewing
image

### When no on is reviewing 
image

### Review page
image


src/main/webapp/app/pages/curation/geneticTypeTabs/GeneticTypeTabs.tsx
```diff
@@ -0,0 +1,141 @@
...
+        <DefaultBadge square color={'warning'} className={sharedClassname} style={sharedStyle}>
+          Pending Release
+        </DefaultBadge>,
+      );
+    }
+
+    return <>{badges.map(badge => badge)}</>;
+  };
+
+  async function handleToggle() {
```
calvinlu3, line 98: Method copied over from SomaticGermlineToggleButton

src/main/webapp/app/pages/curation/geneticTypeTabs/genetic-type-tabs.module.scss
```diff
@@ -0,0 +1,36 @@
+@import '../../../variables.scss';
```
calvinlu3, line 1: Copied from oncokb-public

zhx828:
- [ ] It shows this status. If I truly was in the review page, should I have an option to go to review page when I'm the reviewer?
Screenshot 2024-11-08 at 6 13 32 PM
- [ ] The badge should use the same font as other components
Screenshot 2024-11-08 at 6 16 37 PM

zhx828 (approved):
I can't find any issue with the current code. Once you make additional changes from the comment above. It's good to go.

src/main/webapp/app/pages/curation/geneticTypeTabs/GeneticTypeTabs.tsx
```diff
@@ -0,0 +1,129 @@
...
+      {[GENETIC_TYPE.SOMATIC, GENETIC_TYPE.GERMLINE].map((geneOrigin, idx) => (
+        <div
+          key={idx}
+          style={{ width: '50%' }}
+          className={
+            selected === geneOrigin ? classnames(styles.tab, styles.selectedTab, 'fw-bold') : classnames(styles.tab, styles.unselectedTab)
+          }
+          onClick={handleToggle}
+        >
+          {geneOrigin.substring(0, 1).toUpperCase()}
```
bprize15, line 113: Think it may be simpler here to just hard code the values, especially since it's only two items (and as far as I know it will only ever be two). However, this preference, so feel free to ignore if you disagree.

bprize15 (approved):
Looks really good, thanks!

src/main/webapp/app/pages/curation/geneticTypeTabs/GeneticTypeTabs.tsx
calvinlu3: Yeah - I will keep it as is because I copied most of this file over from the public repository.

PR #471, opened 2024-11-08
move gene creation logic to gene page

bprize15:


src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -48,6 +50,29 @@ export const CurationPage = (props: ICurationPageProps) => {
...
+              history.push(PAGE_ROUTE.CURATION_SOMATIC);
+            }
+          }
+        }
+        setFirebaseGeneExists(true);
+      }
+    }
+
+    checkIfGeneExists();
+  }, [firebaseGenePath, setFirebaseGeneExists, props.firebaseDb, hugoSymbol, isGermline]);
```
zhx828, line 74: setFirebaseGeneExists as part of dependency?

bprize15: It doesn't actually affect anything (that reference never changes), but it considered best practice to put all reactive values (props, variables, and functions) used by useEffect in the dependency array: https://react.dev/reference/react/useEffect#specifying-reactive-dependencies.

zhx828 (approved):
Thanks!

PR #477, opened 2024-11-19
Fix transcript edit page

calvinlu3:
Fixes https://github.com/oncokb/oncokb-pipeline/issues/613

### DIFF
--- a/src/main/webapp/app/entities/alteration/alteration-update.tsx
+++ b/src/main/webapp/app/entities/alteration/alteration-update.tsx
@@ -12,7 +12,6 @@ import { SaveButton } from 'app/shared/button/SaveButton';
 import GeneSelect, { GeneSelectOption } from 'app/shared/select/GeneSelect';
 import { REFERENCE_GENOME } from 'app/config/constants/constants';
 import { Alteration, Gene } from 'app/shared/api/generated/curation';
-import { IGene } from 'app/shared/model/gene.model';
 
 export interface IAlterationUpdateProps extends StoreProps, RouteComponentProps<{ id: string }> {}
 
@@ -20,7 +19,7 @@ export const AlterationUpdate = (props: IAlterationUpdateProps) => {
   const [isNew] = useState(!props.match.params || !props.match.params.id);
   const [proteinChange, setProteinChange] = useState('');
   const [proteinChangeAlteration, setProteinChangeAlteration] = useState(null);
-  const [selectedGenes, setSelectedGenes] = useState<{ label: string | undefined; value: IGene | undefined }[]>([]);
+  const [selectedGenes, setSelectedGenes] = useState<{ label: string | undefined; value: number | undefined }[]>([]);
 
   const consequences = props.consequences;
   const alterationEntity = props.alterationEntity;
@@ -61,7 +60,7 @@ export const AlterationUpdate = (props: IAlterationUpdateProps) => {
       props.alterationEntity.genes && props.alterationEntity.genes.length > 0
         ? props.alterationEntity.genes.map(gene => ({
             label: gene.hugoSymbol,
-            value: gene,
+            value: gene.id,
           }))
         : [];
     setSelectedGenes(_genes);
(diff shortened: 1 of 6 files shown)


src/main/webapp/app/entities/alteration/alteration-update.tsx
```diff
@@ -60,7 +61,7 @@ export const AlterationUpdate = (props: IAlterationUpdateProps) => {
       props.alterationEntity.genes && props.alterationEntity.genes.length > 0
         ? props.alterationEntity.genes.map(gene => ({
             label: gene.hugoSymbol,
-            value: gene.id,
+            value: gene,
```
bprize15, line 64: I would stick with the id for value. If for some reason the reference changes we could end up in a weird state. If you do need to attach the actual gene, could you do it with another property?

src/main/webapp/app/entities/transcript/transcript-update.tsx
```diff
@@ -46,6 +44,22 @@ export const TranscriptUpdate = (props: ITranscriptUpdateProps) => {
...
+    if (transcriptEntity?.gene) {
+      setGeneValue({
+        value: transcriptEntity.gene,
+        label: transcriptEntity.gene.hugoSymbol,
+        synonyms: transcriptEntity.gene.synonyms ?? [],
+      });
+    }
+    if (transcriptEntity?.ensemblGene) {
+      setEnsemblGeneValue({
+        value: transcriptEntity.ensemblGene,
```
bprize15, line 57: Same comment as above about using an id as value

src/main/webapp/app/shared/select/EnsemblGeneSelect.tsx
```diff
@@ -0,0 +1,78 @@
...
+    if (searchWord) {
+      result = await searchEnsemblGenes({ query: searchWord, page: page - 1, size: ITEMS_PER_PAGE, sort: [sortParameter] });
+    } else {
+      result = await getEnsemblGenes({ page: page - 1, size: ITEMS_PER_PAGE, sort: [sortParameter] });
+    }
+
+    options =
+      result?.data?.map(
+        (entity: IEnsemblGene): EnsemblGeneSelectOption => ({
+          value: entity,
```
bprize15, line 43: Same comment as above, prefer id as value

src/main/webapp/app/shared/select/GeneSelect.tsx
```diff
@@ -16,7 +16,7 @@ export interface IGeneSelectProps<IsMulti extends boolean> extends SelectProps<G
 const sortParameter = getEntityPaginationSortParameter(DEFAULT_ENTITY_SORT_FIELD[ENTITY_TYPE.GENE] ?? '', DEFAULT_SORT_DIRECTION);
 
 export interface GeneSelectOption {
-  value?: number;
+  value?: IGene;
```
bprize15, line 19: Same comment

src/main/webapp/app/entities/transcript/transcript-update.tsx
```diff
@@ -54,9 +68,8 @@ export const TranscriptUpdate = (props: ITranscriptUpdateProps) => {
       ...transcriptEntity,
       ...values,
       flags: mapIdList(values.flags),
-      ensemblGene: ensemblGenes.find(it => it.id.toString() === values.ensemblGeneId.toString()),
-      // TYPE-ISSUE: Is selectedGeneId a number or a string?
-      gene: genes.find(it => it.id.toString() === selectedGeneId?.toString()),
+      ensemblGene: ensemblGeneValue ? { id: ensemblGeneValue.value } : undefined,
+      gene: geneValue ? { id: geneValue.value } : undefined,
```
calvinlu3, line 72: @bprize15 Addressed your comments. I actually realized that we don't need to pass in the entire gene object eventhough TranscriptDTO is asking for a `Gene` type. You can just pass in an object with id set to the gene id. 
Actually makes everything simpler.

bprize15 (approved):
LGTM

PR #478, opened 2024-11-19
Allow user to unlock review mode

calvinlu3:
### Issue
Some reviewers may leave the review page open even though they are away from keyboard for a while. This makes it so that the review page is locked.

### Changes
Allow another user to force the current reviewer out, but give a warning.

image


zhx828 (approved):
LGTM

PR #480, opened 2024-11-21
Add keycloak idp mapper for picture attribute

calvinlu3:
In order for the picture from google to be mapped properly, we have to add a mapper in Keycloak.
This PR updates the realm-config json to reflect this change. Profile image is now visible on production curation website (after re-logging)

@bprize15 Could you verify that picture shows up fine for you?


bprize15 (approved):
LGTM

PR #481, opened 2024-11-27
Add readOnly in curation page

gebbing12:
Fixes https://github.com/oncokb/oncokb-pipeline/issues/236

### DIFF
--- a/src/main/webapp/app/pages/curation/CurationPage.tsx
+++ b/src/main/webapp/app/pages/curation/CurationPage.tsx
@@ -124,13 +124,13 @@ export const CurationPage = (props: ICurationPageProps) => {
     firebaseGeneExists &&
     hugoSymbol ? (
     <>
-      {props.readOnly && <ReadOnlyBanner hugoSymbol={hugoSymbol} />}
       <div style={{ visibility: mutationListRendered ? 'visible' : 'hidden' }}>
         <GeneHeader firebaseGenePath={firebaseGenePath} geneEntity={geneEntity} isReviewing={false} />
         <GeneticTypeTabs geneEntity={geneEntity} geneticType={isGermline ? GENETIC_TYPE.GERMLINE : GENETIC_TYPE.SOMATIC} />
         <div className="d-flex justify-content-end mt-2 mb-2">
           <GeneticTypeTabHeader hugoSymbol={hugoSymbol} isReviewing={false} />
         </div>
+        {props.readOnly && <ReadOnlyBanner hugoSymbol={hugoSymbol} />}
         <div className="mb-4">
           <Row className={'justify-content-between'}>
             <Col className="pb-2">
--- a/src/main/webapp/app/pages/curation/CurationPage.tsx
+++ b/src/main/webapp/app/pages/curation/CurationPage.tsx
@@ -25,10 +25,11 @@ import { notifyError } from 'app/oncokb-commons/components/util/NotificationUtil
 import LoadingIndicator, { LoaderSize } from 'app/oncokb-commons/components/loadingIndicator/LoadingIndicator';
 import { parseHistory } from 'app/shared/util/firebase/firebase-history-utils';
 import { useMatchGeneEntity } from 'app/hooks/useMatchGeneEntity';
-import { Unsubscribe, get, ref } from 'firebase/database';
+import { Unsubscribe, get, ref, onValue } from 'firebase/database';
 import { getLocationIdentifier, getTooltipHistoryList } from 'app/components/geneHistoryTooltip/gene-history-tooltip-utils';
 import GeneticTypeTabs, { GENETIC_TYPE } from './geneticTypeTabs/GeneticTypeTabs';
 import GeneticTypeTabHeader from './header/GeneticTypeTabHeader';
+import FlagStore from 'app/entities/flag/flag.store';
 
 export interface ICurationPageProps extends StoreProps, RouteComponentProps<{ hugoSymbol: string }> {}
 
@@ -86,6 +87,11 @@ export const CurationPage = (props: ICurationPageProps) => {
     }
     if (geneEntity && props.firebaseInitSuccess) {
       const cleanupCallbacks: Unsubscribe[] = [];
+      cleanupCallbacks.push(
+        onValue(ref(props.firebaseDb, firebaseMetaCurrentReviewerPath), snapshot => {
+          props.setReadOnly(snapshot.val() ? true : false);
+        }),
+      );
       cleanupCallbacks.push(props.addHistoryListener(firebaseHistoryPath));
       cleanupCallbacks.push(props.addMutationListListener(mutationsPath));
       return () => {
@@ -127,6 +133,7 @@ export const CurationPage = (props: ICurationPageProps) => {
           <Row className={'justify-content-between'}>
             <Col className="pb-2">
               <RealtimeCheckedInputGroup
+                disabled={props.readOnly}
                 groupHeader={
                   <>
                     <span className="me-2">Gene Type</span>
@@ -147,6 +154,7 @@ export const CurationPage = (props: ICurationPageProps) => {
                 })}
               />
               <RealtimeTextAreaInput
+                disabled={props.readOnly}
                 firebasePath={`${firebaseGenePath}/summary`}
                 inputClass={styles.textarea}
                 label="Gene Summary"
@@ -169,6 +177,7 @@ export const CurationPage = (props: ICurationPageProps) => {
           <Row className="mb-3">
             <Col>
               <RealtimeTextAreaInput
+                disabled={props.readOnly}
                 firebasePath={`${firebaseGenePath}/background`}
                 inputClass={styles.textarea}
                 label="Background"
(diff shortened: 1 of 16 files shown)


src/main/webapp/app/pages/curation/collapsible/CancerTypeCollapsible.tsx
```diff
@@ -204,19 +211,21 @@ function CancerTypeCollapsible({
           action={
             <>
               <CommentIcon id={`${cancerTypesUuid}_diagnostic_comments`} path={`${cancerTypePath}/diagnostic_comments`} />
-              <RCTButton cancerTypePath={cancerTypePath} relevantCancerTypesInfoPath={`${cancerTypePath}/diagnostic`} />
+              <RCTButton cancerTypePath={cancerTypePath} relevantCancerTypesInfoPath={`${cancerTypePath}/diagnostic`} readOnly={readOnly} />
```
calvinlu3, line 214: Passing in the readOnly props explicitly is not required because it will be injected by mobx.

calvinlu3 (changes requested):
Thanks for making the changes!

Would you also try adding a banner to the curation page that alerts tells the user that the page is currently in read only mode?
Some text along the lines of "This page is currently in read-only mode because ___ user is currently reviewing"

src/main/webapp/app/pages/curation/button/RCTButton.tsx
```diff
@@ -13,9 +13,16 @@ import { Button } from 'reactstrap';
 export interface IRCTButtonProps extends StoreProps {
   cancerTypePath: string;
   relevantCancerTypesInfoPath: string; // path to dx, px, or tx
+  readOnly?: boolean;
```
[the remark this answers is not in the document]
gebbing12: I added the readOnly property to RCTButton because I thought this component might be used in multiple places simultaneously. If it’s only used in CurationPage, I’ll remove this line and use StoreProp instead.

gebbing12:
> Thanks for making the changes!
> 
> Would you also try adding a banner to the curation page that alerts the user that the page is currently in read only mode? Some text along the lines of "This page is currently in read-only mode because ___ user is currently reviewing"

I have added a banner to the top of the page. Please let me know if there are any improvements needed for the text or style.
691733353841_ pic
ow

calvinlu3:
> I have added a banner to the top of the page. Please let me know if there are any improvements needed for the text or style. 691733353841_ pic

Let's get rid of the "x" button - the alert should not be dismissible.
Also move the alert under the GeneticTypeTabs because the user should be able to edit the germline if somatic is in review (and vice versa).
image

gebbing12:
> > I have added a banner to the top of the page. Please let me know if there are any improvements needed for the text or style. 691733353841_ pic
> 
> Let's get rid of the "x" button - the alert should not be dismissible. Also move the alert under the GeneticTypeTabs because the user should be able to edit the germline if somatic is in review (and vice versa). image



> > I have added a banner to the top of the page. Please let me know if there are any improvements needed for the text or style. 691733353841_ pic
> 
> Let's get rid of the "x" button - the alert should not be dismissible. Also move the alert under the GeneticTypeTabs because the user should be able to edit the germline if somatic is in review (and vice versa). image

Update of banner
WeChatcbf7a0093376990bca3ecb126bdd317a

src/main/webapp/app/pages/curation/header/ReadOnlyBanner.tsx
```diff
@@ -26,15 +26,13 @@ const ReadOnlyBanner = ({ isGermline, hugoSymbol, firebaseDb }: IReadOnlyBanner)
...
-
   return meta && isVisible ? (
     <div className="alert alert-primary alert-dismissible fade show d-flex align-items-center" role="alert">
       <BiInfoCircle size={25} className="me-2" />
-      <span>This page is currently in read-only mode because {meta.review.currentReviewer} is currently reviewing.</span>
-      <button type="button" className="btn-close" aria-label="Close" onClick={handleClose}></button>
+      <span>
+        {isGermline ? 'This Germline ' : 'This Somatic '}
+        is currently in read only mode because {meta.review.currentReviewer} is currently reviewing.
+      </span>
```
calvinlu3, line 35: Maybe `The germline section is current in read only mode because {meta.review.currentReviewer} is currently reviewing.`

src/main/webapp/app/pages/curation/CurationPage.tsx
```diff
@@ -86,6 +88,11 @@ export const CurationPage = (props: ICurationPageProps) => {
     }
     if (geneEntity && props.firebaseInitSuccess) {
       const cleanupCallbacks: Unsubscribe[] = [];
+      cleanupCallbacks.push(
+        onValue(ref(props.firebaseDb, firebaseMetaCurrentReviewerPath), snapshot => {
+          props.setReadOnly(snapshot.val() ? true : false);
```
calvinlu3, line 93: Can keep it concise using double negation
```suggestion
          props.setReadOnly(!!snapshot.val());
```

src/main/webapp/app/pages/curation/button/AddMutationButton.tsx
```diff
@@ -7,9 +7,11 @@ const AddMutationButton: React.FunctionComponent<{
   onClickHandler: (show: boolean) => void;
   showIcon?: boolean;
   showFullTitle?: boolean;
-}> = ({ showAddMutationModal, onClickHandler, showIcon = true, showFullTitle = false }) => {
+  disabled?: boolean;
```
calvinlu3, line 10: I would keep it consistent with the `RCTButton` and just inject the CurationPage store into `AddMutationButton` and use the `readOnly` property for disabled. Same end result, but one less prop to pass in.

src/main/webapp/app/pages/curation/header/ReadOnlyBanner.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { BiInfoCircle } from 'react-icons/bi';
+
+export interface IReadOnlyBanner extends StoreProps {
+  hugoSymbol: string;
+}
+
+const ReadOnlyBanner = ({ isGermline, hugoSymbol, firebaseDb }: IReadOnlyBanner) => {
+  const firebaseMetaPath = getFirebaseMetaGenePath(isGermline, hugoSymbol);
+  const [meta, setMeta] = useState<Meta>();
+  const [isVisible, setIsVisible] = useState(true);
```
calvinlu3, line 17: Seems like this is not needed anymore because we always display the banner.

src/main/webapp/app/stores/curation-page.store.ts
```diff
@@ -133,6 +137,10 @@ export class CurationPageStore {
       .filter(query => query.queryId && this.annotatedAltsCache.cache[query.queryId] && this.annotatedAltsCache.cache[query.queryId].result)
       .map(query => (query.queryId !== undefined ? this.annotatedAltsCache.cache[query.queryId].result : undefined));
   }
+
+  setReadOnly(isAnotherUserReviewing: boolean) {
+    this.readOnly = isAnotherUserReviewing ? true : false;
```
calvinlu3, line 142: Already a boolean value, can just set as is.
```suggestion
    this.readOnly = isAnotherUserReviewing;
```

calvinlu3 (changes requested):
Looks good overall, just some minor comments. Thanks!

src/main/webapp/app/pages/curation/header/ReadOnlyBanner.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { onValue, ref } from 'firebase/database';
+import { getFirebaseMetaGenePath } from 'app/shared/util/firebase/firebase-utils';
+import { BiInfoCircle } from 'react-icons/bi';
+
+export interface IReadOnlyBanner extends StoreProps {
+  hugoSymbol: string;
+}
+
+const ReadOnlyBanner = ({ isGermline, hugoSymbol, firebaseDb }: IReadOnlyBanner) => {
+  const firebaseMetaPath = getFirebaseMetaGenePath(isGermline, hugoSymbol);
```
jfkonecn, line 15: Please put this inside the `useEffect` and add `isGermline` and `hugoSymbol` to the dependency array.
If the path changes you'd would want the `useEffect` to be called again.

jfkonecn (changes requested):

src/main/webapp/app/pages/curation/header/ReadOnlyBanner.tsx
```diff
@@ -0,0 +1,48 @@
...
+    }
+    const subscribe = onValue(ref(firebaseDb, firebaseMetaPath), snapshot => {
+      setMeta(snapshot.val());
+    });
+    return () => subscribe?.();
+  }, []);
+
+  return meta ? (
+    <div className="alert alert-primary alert-dismissible fade show d-flex align-items-center" role="alert">
+      <BiInfoCircle size={25} className="me-2" />
```
calvinlu3, line 32:
```suggestion
      <BiInfoCircle size={DEFAULT_ICON_SIZE} className="me-2" />
```

calvinlu3 (approved):
LGTM! One last minor request.

PR #484, opened 2024-12-19
Fix exon annotation not showing

calvinlu3:
https://github.com/oncokb/oncokb-pipeline/issues/630

### DIFF
--- a/src/main/java/org/mskcc/oncokb/curation/util/parser/ProteinChangeParser.java
+++ b/src/main/java/org/mskcc/oncokb/curation/util/parser/ProteinChangeParser.java
@@ -139,11 +139,8 @@ public class ProteinChangeParser {
             Alteration alteration = new Alteration();
             String revisedProteinChange = "";
             MutationConsequence term = UNKNOWN;
-            if (m.group(1) != null && m.group(3) == null) {
-                // we only want to specify reference when it's one position ins/del
-                alteration.setRefResidues(m.group(1).toUpperCase());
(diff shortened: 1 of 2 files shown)


src/main/java/org/mskcc/oncokb/curation/util/parser/ProteinChangeParser.java
```diff
@@ -139,11 +139,8 @@ public static ParsingStatus<Alteration> parseInframe(String proteinChange) {
             Alteration alteration = new Alteration();
             String revisedProteinChange = "";
             MutationConsequence term = UNKNOWN;
-            if (m.group(1) != null && m.group(3) == null) {
-                // we only want to specify reference when it's one position ins/del
-                alteration.setRefResidues(m.group(1).toUpperCase());
-                revisedProteinChange += alteration.getRefResidues();
-            }
```
calvinlu3, line 146: This change requires discussion. Is `Y770_G772dup` or `770_G772dup` valid?

bprize15 (approved):
LGTM, assuming dup thing is figured out

src/main/java/org/mskcc/oncokb/curation/util/parser/ProteinChangeParser.java
calvinlu3: According to hgvs nomenclature, the reference allele can be include for multi position change

jfkonecn (approved):

PR #485, opened 2024-12-23
Create script to backfill all oncokb canonical transcripts

calvinlu3:
Fixes https://github.com/oncokb/oncokb-pipeline/issues/262

Import the cancer gene list from oncokb core, which has all the oncokb preferred transcripts and reference sequence ids. Backfill in the `OncoKB Canonical` transcript flag and also the refseq ids if they don't exist.

For future releases, we will allow the data releaser to mark the chosen transcripts as canonical as part of data release procedure. [Tracked here](https://github.com/oncokb/oncokb-pipeline/issues/634).

### DIFF
--- a/src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
+++ b/src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
@@ -143,16 +143,6 @@ public class OncokbCurationApp {
         }
     }
 
-    @PostConstruct
-    public void importTranscriptFlags() {
-        Collection<String> activeProfiles = Arrays.asList(
-            env.getActiveProfiles().length == 0 ? env.getDefaultProfiles() : env.getActiveProfiles()
-        );
-        if (activeProfiles.contains("transcript-flag-importer")) {
-            canonicalTranscriptFlagImporter.importCanonicalTranscripts();
-        }
-    }
-
     /**
      * Main method, used to run the application.
      *
--- a/src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
+++ b/src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
@@ -16,7 +16,6 @@ import org.apache.commons.lang3.StringUtils;
 import org.mskcc.oncokb.curation.config.CRLFLogConverter;
 import org.mskcc.oncokb.curation.config.application.ApplicationProperties;
 import org.mskcc.oncokb.curation.config.application.FrontendProperties;
-import org.mskcc.oncokb.curation.importer.CanonicalTranscriptFlagImporter;
 import org.mskcc.oncokb.curation.importer.CdxImporter;
 import org.mskcc.oncokb.curation.importer.Importer;
 import org.oncokb.ApiException;
@@ -43,9 +42,6 @@ public class OncokbCurationApp {
 
     @Autowired
     CdxImporter cdxImporter;
-
-    @Autowired
-    CanonicalTranscriptFlagImporter canonicalTranscriptFlagImporter;
 
     private static final Logger log = LoggerFactory.getLogger(OncokbCurationApp.class);
 
--- a/src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
+++ b/src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
@@ -16,6 +16,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.mskcc.oncokb.curation.config.CRLFLogConverter;
 import org.mskcc.oncokb.curation.config.application.ApplicationProperties;
 import org.mskcc.oncokb.curation.config.application.FrontendProperties;
+import org.mskcc.oncokb.curation.importer.CanonicalTranscriptFlagImporter;
 import org.mskcc.oncokb.curation.importer.CdxImporter;
 import org.mskcc.oncokb.curation.importer.Importer;
 import org.oncokb.ApiException;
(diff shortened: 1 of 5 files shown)


src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
```diff
@@ -139,6 +143,16 @@ public void importInitialCdxData() throws IOException {
...
+    @PostConstruct
+    public void importTranscriptFlags() {
+        Collection<String> activeProfiles = Arrays.asList(
+            env.getActiveProfiles().length == 0 ? env.getDefaultProfiles() : env.getActiveProfiles()
+        );
+        if (activeProfiles.contains("transcript-flag-importer")) {
+            canonicalTranscriptFlagImporter.importCanonicalTranscripts();
+        }
+    }
+
```
jfkonecn, line 155: Couldn't you have all this logic inside the importer itself?

src/main/java/org/mskcc/oncokb/curation/importer/CanonicalTranscriptFlagImporter.java
```diff
@@ -0,0 +1,120 @@
...
+            InputStream is = geneListUrl.openStream();
+            List<List<String>> tsvFile = readDelimitedLinesStream(is, "\t", true);
+            for (int i = 0; i < tsvFile.size(); i++) {
+                if (i == 0) {
+                    continue;
+                }
+                List<String> line = tsvFile.get(i);
+                if (line.size() < 8) {
+                    log.warn(String.format("Skipping line %i", i));
+                    continue;
```
jfkonecn, line 61: Is the data corrupted if this happens?

src/main/java/org/mskcc/oncokb/curation/importer/CanonicalTranscriptFlagImporter.java
```diff
@@ -0,0 +1,120 @@
...
+        FlagService flagService,
+        InfoService infoService,
+        MainService mainService
+    ) {
+        this.geneService = geneService;
+        this.transcriptService = transcriptService;
+        this.flagService = flagService;
+    }
+
+    public void importCanonicalTranscripts() {
```
jfkonecn, line 49: Is this basically a data migration? In other words, does this only need to be run with a new data release?

src/main/java/org/mskcc/oncokb/curation/importer/CanonicalTranscriptFlagImporter.java
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * To backfill the OncoKB canonical transcripts
+ */
+
+@Component
+public class CanonicalTranscriptFlagImporter {
+
+    private final Logger log = LoggerFactory.getLogger(CanonicalTranscriptFlagImporter.class);
+
+    final String ONCOKB_GENE_LIST_FILE_PATH = "data/oncokb_core/cancer_gene_list_v4_23.tsv";
```
jfkonecn, line 29: We're on v4.24 for data releases. Did you mean v4.24?

jfkonecn (changes requested):

src/main/java/org/mskcc/oncokb/curation/importer/CanonicalTranscriptFlagImporter.java
calvinlu3: Shouldn't happen unless the cancer gene list file we generated in oncokb-data repo is corrupted.

src/main/java/org/mskcc/oncokb/curation/importer/CanonicalTranscriptFlagImporter.java
calvinlu3: This can be ran with a new data release if needed, but I would just wait for the feature oncokb/oncokb-pipeline#634.
The main purpose is to update the transcript information we have saved in oncokb_curation with the information in oncokb_core.

src/main/java/org/mskcc/oncokb/curation/importer/CanonicalTranscriptFlagImporter.java
jfkonecn: In that case I think it should throw an exception. The file is in source control so there shouldn't be a reason why it would not fail locally, but only in production. If it's a warning it's easy to miss.

src/main/java/org/mskcc/oncokb/curation/importer/CanonicalTranscriptFlagImporter.java
calvinlu3: Yeah good point. I will update with v4.24

src/main/java/org/mskcc/oncokb/curation/OncokbCurationApp.java
calvinlu3: Yeah I don't see why not - could also extract the logic to check if profile is active into a util too. I just copied the cdxImporter.

src/main/java/org/mskcc/oncokb/curation/importer/CanonicalTranscriptFlagImporter.java
```diff
@@ -0,0 +1,129 @@
...
+        this.geneService = geneService;
+        this.transcriptService = transcriptService;
+        this.flagService = flagService;
+    }
+
+    @Override
+    public void run(String... args) throws Exception {
+        log.info("Running canonical transcript flag importer...");
+        importCanonicalTranscripts();
+        log.info("Finished importer.");
```
jfkonecn, line 56: Did you mean to add a guard checking if the correct profile is being used?

jfkonecn (approved):

PR #498, opened 2025-02-14
Fixed global search

jfkonecn:
closes https://github.com/oncokb/oncokb-pipeline/issues/657

Currently we're getting an error `Parameter 1 of function 'upper()' has type 'STRING', but argument is of type 'java.lang.String' mapped to 'CLOB'`

The name column in the drug table is of type `longtext` which apparently is a `CLOB` type so the name column needs to be explicitly casted to a string type.

### DIFF
--- a/src/main/java/org/mskcc/oncokb/curation/service/DrugQueryService.java
+++ b/src/main/java/org/mskcc/oncokb/curation/service/DrugQueryService.java
@@ -1,7 +1,9 @@
 package org.mskcc.oncokb.curation.service;
 
 import jakarta.persistence.criteria.JoinType;
+import java.util.Iterator;
 import java.util.List;
+import java.util.function.Function;
 import org.mskcc.oncokb.curation.domain.*; // for static metamodels
 import org.mskcc.oncokb.curation.domain.Drug;
 import org.mskcc.oncokb.curation.repository.DrugRepository;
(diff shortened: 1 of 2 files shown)


src/main/java/org/mskcc/oncokb/curation/service/DrugQueryService.java
```diff
@@ -96,8 +99,13 @@ protected Specification<Drug> createSpecification(DrugCriteria criteria) {
             if (criteria.getId() != null) {
                 specification = specification.or(buildRangeSpecification(criteria.getId(), Drug_.id));
             }
-            if (criteria.getName() != null) {
-                specification = specification.or(buildStringSpecification(criteria.getName(), Drug_.name));
+            if (criteria.getName() != null && criteria.getName().getContains() != null) {
+                specification = specification.or((root, _ignore, cb) -> {
+                    return cb.like(
+                        cb.lower(root.get(Drug_.name).as(String.class)),
```
calvinlu3, line 105: MySQL `LIKE` operator should already compare ignoring case right? Does that mean we don't need `cb.lower(root.get(Drug_.name).as(String.class))`, instead we can trim to just `root.get(Drug_.name).as(String.class)`?

calvinlu3 (approved):

PR #499, opened 2025-02-17
fix bug where empty treatment caused empty collapsible

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/662

### DIFF
--- a/src/main/webapp/app/shared/table/GenomicIndicatorsTable.tsx
+++ b/src/main/webapp/app/shared/table/GenomicIndicatorsTable.tsx
@@ -221,20 +221,11 @@ const GenomicIndicatorsTable = ({
                   isDisabled={genomicIndicator.name_review?.removed || readOnly || false}
                   value={
                     genomicIndicator.associationVariants?.map(variant => {
-                      // remove when working on https://github.com/oncokb/oncokb-pipeline/issues/389
-                      if (variant.uuid === PATHOGENIC_VARIANTS) {
-                        return { label: PATHOGENIC_VARIANTS, value: variant.uuid };
-                      }
-
                       const associatedMutation = mutations?.find(mutation => mutation.name_uuid === variant.uuid);
                       return { label: getMutationName(associatedMutation?.name, associatedMutation?.alterations), value: variant.uuid };
                     }) || []
                   }
                   options={[
-                    {
-                      label: PATHOGENIC_VARIANTS,
-                      value: PATHOGENIC_VARIANTS,
-                    },
                     ...(mutations?.map(mutation => ({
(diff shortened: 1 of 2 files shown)


src/main/webapp/app/shared/table/GenomicIndicatorsTable.tsx
```diff
@@ -221,20 +221,11 @@ const GenomicIndicatorsTable = ({
...
                       const associatedMutation = mutations?.find(mutation => mutation.name_uuid === variant.uuid);
                       return { label: getMutationName(associatedMutation?.name, associatedMutation?.alterations), value: variant.uuid };
                     }) || []
                   }
                   options={[
-                    {
-                      label: PATHOGENIC_VARIANTS,
-                      value: PATHOGENIC_VARIANTS,
-                    },
                     ...(mutations?.map(mutation => ({
```
jfkonecn, line 229: Why is this getting deleted?

bprize15: First I should mention that this seems outside the scope of the current ticket. I traced the introduction of the bug down to this PR: https://github.com/oncokb/oncokb-transcript/pull/491. When fixing this issue, I wanted to make sure the fix in that ticket was not being overwritten, which caused me to notice this issue as well.

It's not needed anymore. In my original implementation, I always had a placeholder option for Pathogenic variants. However, somewhere along the line the implementation changed so that Pathogenic mutations was created with the introduction of a new genomic indicator, so the default option is no longer necessary. This can even result in a bug where the placeholder option and the created mutation exist, causing Pathogenic mutations to appear twice.

jfkonecn (approved):

calvinlu3 (approved):
Thanks - feel free to make a release so curators can continue reviewing

src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx
```diff
@@ -268,7 +268,9 @@ export const getCompactReviewInfo = <T extends BaseReviewLevel>(review: T): T =>
   }
   let childReview = review.children[0];
   if (childReview?.nestedUnderCreateOrDelete) {
-    return review;
+    if (isCreateReview(review)) {
+      return review;
+    }
   }
```
calvinlu3, line 274: I don't remember exactly why I changed this. However, in that PR that you pointed out I think I wanted to return review for isDeleteReview as well.

calvinlu3: If you cannot reproduce this (https://github.com/oncokb/oncokb-pipeline/issues/643#issuecomment-2617196177) issue, then feel free to merge

PR #509, opened 2025-04-16
Removed quotes from value on data import

jfkonecn:
image

### DIFF
--- a/src/main/webapp/app/pages/curation/list/FirebaseList.tsx
+++ b/src/main/webapp/app/pages/curation/list/FirebaseList.tsx
@@ -71,7 +71,7 @@ function FirebaseList<T>({ path, itemBuilder, pushDirection, scrollOptions, filt
         return;
       }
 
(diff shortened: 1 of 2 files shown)


src/main/webapp/app/pages/curation/list/FirebaseList.tsx
```diff
@@ -66,7 +66,7 @@ function FirebaseList<T>({ path, itemBuilder, pushDirection, scrollOptions, filt
         return;
       }
       const items = (await get(ref(firebaseDb, path))).val();
-      if (!items && indices?.length !== 0) {
+      if (!items || indices?.length !== 0) {
```
calvinlu3, line 69: What's the reason for this change?

jfkonecn: `item` could be null coming out of firebase and it was crashing the page.

jfkonecn: Looks like screenshot tests aren't happy with this code...

PR #511, opened 2025-04-21
do not append null to inframe string

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/733


src/main/java/org/mskcc/oncokb/curation/util/parser/ProteinChangeParser.java
```diff
@@ -142,7 +142,7 @@ public static ParsingStatus<Alteration> parseInframe(String proteinChange) {
                 // we only want to specify reference when it's one position ins/del
                 alteration.setRefResidues(m.group(1).toUpperCase());
             }
-            revisedProteinChange += alteration.getRefResidues();
+            revisedProteinChange += m.group(1);
```
calvinlu3, line 145: So basically, it will just use whatever the user entered as reference allele, right?

bprize15: Yeah exactly

PR #517, opened 2025-04-24
allow download callback for oncokb table

bprize15:
resolves https://github.com/oncokb/oncokb-pipeline/issues/734

@calvinlu3 I originally planned on allowing the table itself to handle the download logic. However, this is complicated since some instances of the table pass in a Cell() prop to access the value. It is not always so simple as using an accessor or ID, so I figured it was best left to the user of the table to implement the download functionality.


calvinlu3 (approved):
Looks good! Thank you

PR #522, opened 2025-05-12
Add excel like column filtering and fix CDx table searching

calvinlu3:
Fixes https://github.com/oncokb/oncokb-pipeline/issues/749

### Changes
- Add excel-like filtering
   - Number, String and Date filters
- Fix CDx search and filtering


src/main/webapp/app/pages/curation/GeneListPage.tsx
```diff
@@ -70,20 +70,21 @@ const GeneListPage = (props: IGeneListPage) => {
...
-      onFilter: (data: GeneMetaInfo, keyword) => (data.hugoSymbol ? filterByKeyword(data.hugoSymbol, keyword) : false),
+      onSearchFilter: (data: GeneMetaInfo, keyword) => (data.hugoSymbol ? filterByKeyword(data.hugoSymbol, keyword) : false),
     },
     {
       accessor: 'lastModifiedAt',
       Header: 'Last modified at',
+      filterType: FilterTypes.DATE,
       getColumnFilterValue(data: GeneMetaInfo) {
-        return data.lastModifiedAt ? moment(parseInt(data.lastModifiedAt, 10)).format(APP_DATETIME_FORMAT) : '';
+        return new Date(parseInt(data.lastModifiedAt, 10));
```
bprize15, line 87: Have we guaranteed last modified at will always be available?

src/main/webapp/app/shared/table/OncoKBTable.tsx
```diff
@@ -37,18 +64,35 @@ export const OncoKBTable = <T extends object>({ disableSearch = false, showPagin
...
+
+        switch (curColumn.filterType) {
+          case FilterTypes.STRING:
+            return typeof value === 'string' && (selectedValues as Set<string>).has(value);
+
+          case FilterTypes.NUMBER:
+            return typeof value === 'number' && [...selectedValues].some(v => Number(v) === value);
+
+          case FilterTypes.DATE:
+            return value instanceof Date && [...selectedValues].some(v => new Date(v).getTime() === value.getTime());
```
bprize15, line 81: Wondering if this filter is hard to use since it is comparing exact millisecond time?

bprize15 (changes requested):
Great feature!

The date filter does not seem to be working as intended. Only the buttons on the left (day, week, etc.) are working correctly for me. Dates outside my range are not being filtered out.

One other nit: The page number is not updated when a filter is applied. For example, if I first navigate to page 7 and then filter by my name (I only have one page worth of data), the page is displayed as 7/1

src/main/webapp/app/pages/curation/GeneListPage.tsx
calvinlu3: Not sure why we had this check before. The firebase model has that field as required, so I removed.

calvinlu3:
> One other nit: The page number is not updated when a filter is applied. For example, if I first navigate to page 7 and then filter by my name (I only have one page worth of data), the page is displayed as 7/1

Oh didn't even notice, thanks!
It seems like its an internal issue with the package and it would require hacky workarounds to get it to reset page when the underly data has been updated. See https://github.com/TanStack/table/  issues/1206
I will skip for now.

bprize15 (approved):
LGTM

PR #532, opened 2025-07-28
Fixed infinite re-rendering when you delete a cancer type

ReshmaRamaiah10:
Issue: An infinite re-render loop was caused when a key was deleted because `addedListItemKeys` became an empty array `[]`.

Fix: Compare the new filtered keys with the current `addedListItemKeys` by checking both length and content equality. Update `addedListItemKeys` only if they are different.


bprize15:
> Issue: An infinite re-render loop was caused when a key was deleted because `addedListItemKeys` became an empty array `[]`.
> 
> Fix: Compare the new filtered keys with the current `addedListItemKeys` by checking both length and content equality. Update `addedListItemKeys` only if they are different.

Could you check out a new branch describing the name of the feature. Something like `fix/infinite-rerender`. Makes it easier since I already have a local copy of the rc branch

PR #533, opened 2025-07-30
Fixed infinite re-rendering when you delete a cancer type

ReshmaRamaiah10:
Issue: An infinite re-render loop was caused when a key was deleted because `addedListItemKeys` became an empty array `[]`.

Fix: Compare the new filtered keys with the current `addedListItemKeys` by checking both length and content equality. Update `addedListItemKeys` only if they are different.

Previous PR: https://github.com/oncokb/oncokb-transcript/pull/532


bprize15 (approved):
LGTM

PR #534, opened 2025-07-31
Fix/gene review false

ReshmaRamaiah10:
https://github.com/oncokb/oncokb-pipeline/issues/829

When a change is made, a `uuid` is added to the review object with a value of `true`. This indicates that a review is required. Once the change is reviewed, the uuid’s value is set to `null`, which triggers the removal of the UUID from the review object.

However, if the change is reverted before being reviewed, `isChangeReverted` is set to `true`. 
`[uuid]: !isChangeReverted` results in `uuid: false`, which prevents the UUID from being removed, causing the system to still show that a review is required.

**Fix:** If `isChangeReverted` is `true`, the `uuid` is set to null
`[uuid]: !isChangeReverted ? true : null`


src/main/webapp/app/pages/curation/collapsible/Collapsible.tsx
```diff
@@ -25,8 +25,8 @@ export default function Collapsible({
 }: CollapsibleProps) {
   const defaultColorOptions: CollapsibleColorProps | undefined = colorOptions
     ? {
-        hideLeftBorder: colorOptions?.hideLeftBorder ? colorOptions.hideLeftBorder : false,
         ...colorOptions,
+        hideLeftBorder: colorOptions.hideLeftBorder ?? false,
```
ReshmaRamaiah10, line 29: John suggested the change because:
1. We already check if `colorOptions` is `undefined`
2. moving `hideLeftBorder` logic under `...colorOptions` ensures its overridden with the correct value

src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
```diff
@@ -98,7 +98,9 @@ export class FirebaseGeneReviewService {
     const { hugoSymbol } = parseFirebaseGenePath(firebasePath) ?? {};
 
     let updateObject = this.getGeneUpdateObject(updateValue, updatedReview!, firebasePath, uuid!);
-    const metaUpdateObject = this.firebaseMetaService.getUpdateObject(hugoSymbol!, isGermline, { [uuid!]: !isChangeReverted });
+    const metaUpdateObject = this.firebaseMetaService.getUpdateObject(hugoSymbol!, isGermline, {
```
bprize15, line 101: Makes sense to me, but would you mind just checking with Calvin? Feels like setting `!isChangeReverted` and never expecting the value to be false is a weird oversight. Figuring Calvin may have had a use case in mind

calvinlu3: We can go ahead with this change. If a change was reverted, we do want to remove the uuid from the meta collection because it no longer needs review. Fine with this change

src/main/webapp/app/service/firebase/firebase-gene-review-service.ts
```diff
@@ -98,7 +98,9 @@ export class FirebaseGeneReviewService {
...
+      [uuid!]: !isChangeReverted ? true : null,
```
calvinlu3, line 102: We should really check if uuid is undefined/null and then generate one if it does not exist instead of using `[uuid!]`, this will cause `null: true` to be inserted.

ReshmaRamaiah10: Is` if (uuid != null)` enough? I think I’ve seen` null: true` used somewhere before, not sure if it was from legacy platform.

PR #586, opened 2026-02-20
Fix websocket unauthenticated error

calvinlu3:
resolves https://github.com/oncokb/oncokb-pipeline/issues/1113

The WebSocket handshake to `/websocket/curation/validation` was authenticated (OAuth2 token present in the session) and included the correct roles but still returned `403` because the req arrived with `X-Forwarded-Proto: wss`, while the browser origin was https://beta.curation.oncokb.org.
Since no allowed origins were configured for WebSockets, spring treated this as cross origin and rejected the handshake.

After the nginxingress issuue on Feb 12, traffic started flowing through Traefik.
Traefik’s forwarded headers caused the backend to see `X‑Forwarded‑Proto: wss` for the WebSocket handshake.
With nginx ingress, the backend typically saw https so the Origin check passed without explicit configuration.


jfkonecn (approved):
Weird bug. Nice work tracking it down!

PR #587, opened 2026-02-25
Fix/drug review

calvinlu3:
resolves https://github.com/oncokb/oncokb-pipeline/issues/1122

In Review mode, treatment diffs display raw UUIDs instead of drug names because the path regex in ReviewCollapsible.tsx only matches numeric array keys (`treatments/\\d+/name`). After the Firebase arrays to maps change, treatment paths use UUID keys, so the regex doesn’t match and getTxName(...) isn’t applied.


calvinlu3:
@ReshmaRamaiah10 Take your time to understand the issue. I will merge now so that Sarah can continue reviewing for data release

PR #590, opened 2026-03-19
Keep review page in sync with Firebase during review

jfkonecn:
Closes https://github.com/oncokb/oncokb-pipeline/issues/1116

AI Generated Summary:

Rejecting one review item could leave stale review state in ReviewPage, so later accepts were still operating on an outdated root review tree. In the repro case, rejecting the description and then accepting oncogenicity/pathogenicity caused the old description data to remain on the curation page because the review UI was not re-synced from Firebase after those changes.

This change switches the review page from one-time Firebase reads to live listeners for both gene data and review metadata. That keeps the computed root review tree current as review actions update Firebase, so child components rerender against fresh state and rejected fields do not get reintroduced by subsequent review actions. It also keeps the accept-all flow aligned with the listener-based refresh model and softens the staging save failure path so local review work is not interrupted by staging connectivity issues.

### DIFF
--- a/src/main/webapp/app/pages/curation/review/ReviewPage.tsx
+++ b/src/main/webapp/app/pages/curation/review/ReviewPage.tsx
@@ -10,7 +10,7 @@ import { getFirebaseGenePath, getFirebaseMetaGenePath } from 'app/shared/util/fi
 import { componentInject } from 'app/shared/util/typed-inject';
 import { getSectionClassName, useDrugListRef } from 'app/shared/util/utils';
 import { IRootStore } from 'app/stores';
-import { get, ref, set } from 'firebase/database';
+import { onValue, ref, set } from 'firebase/database';
 import { observer } from 'mobx-react';
 import React, { useEffect, useState } from 'react';
 import { Alert, Col, Row } from 'reactstrap';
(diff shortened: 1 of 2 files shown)


src/main/webapp/app/pages/curation/review/ReviewPage.tsx
```diff
@@ -51,23 +51,30 @@ const ReviewPage: React.FunctionComponent<IReviewPageProps> = (props: IReviewPag
...
+    if (!geneEntity || !props.firebaseInitSuccess || !props.firebaseDb || !hugoSymbol) {
       return;
     }
-    // Fetch the data when the user enters review mode. We don't use a listener
-    // because there shouldn't be another user editing the gene when it is being reviewed.
-    const geneDataSnapshot = await get(ref(props.firebaseDb, firebaseGenePath));
-    setGeneData(geneDataSnapshot.val());
-    const metaReviewSnapshot = await get(ref(props.firebaseDb, firebaseMetaReviewPath));
-    setMetaReview(metaReviewSnapshot.val());
-  };
```
calvinlu3, line 64: Is there a way to fix this bug without removing this? One of the reasons I can remember is that we don't want the screen to have so many moving parts when the listener gets triggered. (ie the collapsibles start to get condensed after accepting/rejecting something)

calvinlu3: I think an option can be that when we call `getCreateUpdateObject(..., reviewLevel,...)`, we can choose to not use the reviewLevel (which was generated once when user enters review mode) and choose to fetch the latest data for only that path from firebase.

calvinlu3 (changes requested):

src/main/webapp/app/pages/curation/review/ReviewPage.tsx
jfkonecn: Is there a way we can prove that having the listener is or is not a problem?

calvinlu3: Maybe let's have a call about this, might be easier to explain.

jfkonecn:
I made a new PR
https://github.com/oncokb/oncokb-transcript/pull/593

PR #591, opened 2026-03-20
Feature/tt specific desc

calvinlu3:
related to https://github.com/oncokb/oncokb-pipeline/issues/1096


src/main/webapp/app/pages/curation/collapsible/TherapyTxDescAddendumsSection.tsx
```diff
@@ -0,0 +1,424 @@
...
+  firebaseDb,
+  deleteSection,
+  updateReviewableContent,
+}: ITherapyTxDescAddendumsSectionProps) {
+  const [txDescAddendums, setTxDescAddendums] = useState<TxDescAddendumList>({});
+  const [cancerTypes, setCancerTypes] = useState<CancerTypeList>({});
+  const [excludedCancerTypes, setExcludedCancerTypes] = useState<CancerTypeList>({});
+  const [selectableCancerTypes, setSelectableCancerTypes] = useState<CancerType[]>([]);
+  const [allCancerTypeOptions, setAllCancerTypeOptions] = useState<CancerTypeSelectOption[]>([]);
+
```
bprize15, line 59: Assuming the curators do not require this feature for "Other Tumor Types"?

bprize15:
Product Review: When I add an addendum and try to remove it immediately, it goes into review mode instead of deleting immediately. This creates a weird cascade of bugs, where it shows updates, and in testing, it is showing firebase list keys. 

Screenshot 2026-03-30 at 4 09 53 PM

I am having trouble reproducing the one that makes the firebase list key shown up, so it is possible that has to do with my data. But the review mode bug is easily reproducible

bprize15 (changes requested):

src/main/webapp/app/pages/curation/collapsible/TherapyTxDescAddendumsSection.tsx
calvinlu3: Yup, currently restricting it to only All Solid/Liquid Tumors. Other Tumor Types would be interesting because if they wanted a specific "other" tumor type addendum, then they should probably curate that specific tumor type explicitly. They only want it for these two special tt for now.

calvinlu3:
> Product Review: When I add an addendum and try to remove it immediately, it goes into review mode instead of deleting immediately. This creates a weird cascade of bugs, where it shows updates, and in testing, it is showing firebase list keys.

@bprize15 Thank you very much for the product review! I was able to reproduce the bug where we cannot delete immediately. That is fixed now.

src/main/webapp/app/shared/select/CancerTypeSelect.tsx
```diff
@@ -127,7 +127,6 @@ const CancerTypeSelect = <IsMulti extends boolean>(props: ICancerTypeSelectProps
       }
       cacheUniqs={[props.value]}
       placeholder="Select cancer type"
-      isClearable
```
bprize15, line 130: Do we want all cancer type selects to not be clearable?

src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx
```diff
@@ -488,6 +556,19 @@ export const findReviewRecursive = (
       if (key.endsWith('_uuid') && uuids.includes(value as string)) {
         const relevantKeys = getRelevantKeysFromUuidKey(key);
 
+        if (relevantKeys.fieldKey === 'cancer_type') {
```
bprize15: Is this assuming that all other objects that contain a cancer type have already been processed?

src/main/java/org/mskcc/oncokb/curation/config/security/SecurityConfigurationOAuth.java
```diff
@@ -59,6 +60,7 @@ public SecurityFilterChain oauthFilterChain(HttpSecurity http, MvcRequestMatcher
                         .csrfTokenRequestHandler(new SpaCsrfTokenRequestHandler())
             )
             .addFilterAfter(new SpaWebFilter(), BasicAuthenticationFilter.class)
+            .addFilterAfter(new WebsocketAuthDebugFilter(), BasicAuthenticationFilter.class)
```
bprize15: Based on the code, I don't see a reason you need to omit this in production, but since you call it a debug filter, is it intended to only log when the application is run in debug mode? If so, maybe we want the log level to be DEBUG. No preference here.

src/main/webapp/app/pages/curation/collapsible/TherapyTxDescAddendumsSection.tsx
```diff
@@ -0,0 +1,417 @@
...
+  useEffect(() => {
+    if (!showAddTumorTypeSpecificDescriptionButton || Object.keys(txDescAddendums).length === 0) {
+      setAllCancerTypeOptions([]);
+      return;
+    }
+
+    const pageSize = 1000;
+
+    const fetchAllCancerTypes = async () => {
+      try {
```
bprize15, line 186: This logic seems a bit complex to get all cancer types. Would it be easy to use a non-paginated version?

bprize15 (commented):
One more product comment that I just noticed. You can add a description without a cancer type an approve it. This might cause unexpected data for the data release CLI or cause us to handle a case that does not make sense. Maybe we want to enforce this constraint here so we cannot have invalid data in our system.

src/main/webapp/app/shared/select/CancerTypeSelect.tsx
calvinlu3: It used to override the prop, so it was always clearable. I double checked all places we used CancerTypeSelect and passed in the prop if we really needed it to be clearable. The default behavior if select is Multi, is that it will be clearable.

src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx
calvinlu3: Sorry for the confusion here. I ended up refactoring the whole tx addendum review building. Now its just `buildTxDescAddendumReview`.

src/main/java/org/mskcc/oncokb/curation/config/security/SecurityConfigurationOAuth.java
calvinlu3: Not sure how this code got into my branch, but it was for testing. I got rid of the associated file and this line.

src/main/webapp/app/pages/curation/collapsible/TherapyTxDescAddendumsSection.tsx
calvinlu3: Unfortunately, this is a big piece of code to do a simple thing because
the codegen generated api.ts does not serialize the params properly. For some reason it serializes Pageable into `<Object>` instead of adding each of the Pageable field (sort,size,page) into url params. I added a comment so we can try to fix it later.

I'm going to skip this for now, but we can think about whether we want to support non-paginated endpoint or have some reusable logic on frontend to get all with paginated version. Also, if we even need pagination for cancer type since theres so few rows.

calvinlu3:
> One more product comment that I just noticed. You can add a description without a cancer type an approve it. This might cause unexpected data for the data release CLI or cause us to handle a case that does not make sense. Maybe we want to enforce this constraint here so we cannot have invalid data in our system.

The description is disabled until a cancer type is added. Also, the data release cli will generate an error if either are missing.

bprize15 (approved):
LGTM

PR #593, opened 2026-04-20
Fix stale entity writes when accepting create reviews

jfkonecn:
closes https://github.com/oncokb/oncokb-pipeline/issues/1116

AI Summary:

The core production fix is in src/main/webapp/app/service/firebase/firebase-gene-review-service.ts:246. Before, accepting a create review wrote reviewLevel.historyData.newState back to Firebase. The unstaged change now:

- derives createdEntityPath
- fetches the current entity from Firebase with firebaseRepository.get(...)
- throws if that entity is missing
- uses the fresh entity for both history (historyReviewLevel.historyData.newState = currentCreatedEntity) and the final write
- clears nested review flags on a clone of that fresh entity before saving

That is exactly the behavior described in review-create-accept-stale-data-fix.md.

The regression test in src/main/webapp/app/service/firebase/firebase-gene-review-service.spec.ts:356 is also directly tied to the doc. It creates the stale-vs-latest scenario the note describes:

- review tree contains "Stale summary"
- mocked Firebase returns "Latest summary"
- assertion verifies the update writes the latest Firebase entity and does not write the stale summary

There is also a small test adjustment just above it in the existing accept-create test at src/main/webapp/app/service/firebase/firebase-gene-review-service.spec.ts:332, where mockFirebaseRepository.get is now set up because the production code now reads from Firebase during accept.

Related nearby UI/tree changes, but not required for the fix

These match the note’s “separate” UI/display changes.

In src/main/webapp/app/shared/util/firebase/firebase-review-utils.tsx:289, a new compactReviewTree(...) helper recursively compacts the whole review tree instead of only compacting one level at a time. Its test was added in src/main/webapp/app/shared/util/firebase/firebase-review-utils.spec.ts:667.

In src/main/webapp/app/pages/curation/review/ReviewPage.tsx:92, the page now calls compactReviewTree(reviews) once when building the tree, instead of compacting only the top-level children. It also removes a setRootReview(_.cloneDeep(rootReview)) refresh in rootDelete at src/main/webapp/app/pages/curation/review/ReviewPage.tsx:273, so the page no longer forces a full root rerender while actions are pending.

In src/main/webapp/app/pages/curation/collapsible/ReviewCollapsible.tsx:106, the component stops re-compacting children locally on mount, keeps a separate reviewChildren state, and renders from that state at src/main/webapp/app/pages/curation/collapsible/ReviewCollapsible.tsx:393. That makes the optimistic hide/delete behavior local to the collapsible instead of depending on a root-level clone refresh.

So, if you want the shortest answer: the stale-data bug fix is the service change plus its tests; the util/page/collapsible diffs are adjacent refactors for review-tree compaction and UI state handling, not the minimal fix itself.


src/test/javascript/spec/screenshot.ts
jfkonecn, line 1: For some reason the screenshot tests stopped working so codex added a bunch of waits to fix it.

calvinlu3 (approved):
Thanks!

PR #601, opened 2026-06-01
Investigate/create existing gene

calvinlu3:
fixes https://github.com/oncokb/oncokb-pipeline/issues/1216

The reason for this issue was that when user was on
1. Germline toggle on gene list page
2. Try to add a gene that is in Somatic genes collection, but not in Germline_Genes collection
3. Allows you to overwrite the somatic gene

## Fix
- Guard against overwriting when creating (introduced createIfAbsent)
- Do not allow creating new gene if it exists in somatic.


jfkonecn (approved):

calvinlu3:
One of the webdriver test passed, so it looks like the failing one ran into a timing issue with webdriver.

PR #603, opened 2026-06-04
Fix placement of ref in editor

calvinlu3:
fixes https://github.com/oncokb/oncokb-pipeline/issues/1050

Issue was that when a user places their cursor in the middle of the text and goes to add a reference, the reference ends up getting added to the end of the text
Reason was that the editor's cursor pos was not stored and clicking on the ref button in the toolbar unfocuses the editor and the pos is lost.

### DIFF
--- a/src/main/webapp/app/shared/rich-text-editor/addons/ReferenceToolbarAddon.tsx
+++ b/src/main/webapp/app/shared/rich-text-editor/addons/ReferenceToolbarAddon.tsx
@@ -1,7 +1,9 @@
 import classNames from 'classnames';
+import { Editor } from '@tiptap/core';
 import React, { useState } from 'react';
 import * as styles from '../RichTextEditor.module.scss';
-import { useRichTextEditorContext } from '../RichTextEditor';
+import { RichTextSelectionRange, useRichTextEditorContext } from '../RichTextEditor';
+import { RichTextInlineNode } from '../richTextSchema';
 import { isSafeHttpUrl, normalizeHttpUrl } from '../utils';
 import { buildAbstractReferenceNode, buildNctReferenceNode } from './ReferenceNodeExtensions';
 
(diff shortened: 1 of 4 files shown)


src/main/webapp/app/shared/rich-text-editor/addons/ReferenceToolbarAddon.tsx
```diff
@@ -122,7 +129,12 @@ export const AbstractToolbarAddon: React.FC = () => {
     const title = abstractTitle.trim();
     const href = normalizeHttpUrl(abstractLink);
     if (!title || !href || !editor || !isSafeHttpUrl(href)) return;
-    editor.chain().focus().insertContent(buildAbstractReferenceNode(title, href)).run();
+    const selection = getSavedEditorSelection();
```
bprize15: Nit: maybe we could use a helper method and pass a callback to `insertContent`

src/main/webapp/app/shared/rich-text-editor/addons/ReferenceToolbarAddon.tsx
```diff
@@ -75,14 +77,19 @@ export const PmidToolbarAddon: React.FC = () => {
...
   const isOpen = activeToolbarPopover === NCT_POPOVER_ID;
 
   const insertNct = () => {
     const nctId = nctInput.trim();
     if (!nctId || !editor) return;
-    editor.chain().focus().insertContent(buildNctReferenceNode(nctId)).run();
+    const selection = getSavedEditorSelection();
+    const chain = editor.chain().focus();
+    if (selection) {
+      chain.setTextSelection(selection.from);
```
bprize15: I think I would expect that if `from != to` we would want to overwrite the highlighted text rather than insert at the start. But if you disagree, feel free to leave as is.

bprize15 (approved):
LGTM with minor comments

PR #604, opened 2026-06-24
Added new transcript endpoint for data-release

jfkonecn:


src/main/java/org/mskcc/oncokb/transcript/service/GeneService.java
```diff
@@ -146,6 +152,62 @@ public Optional<Gene> findGeneByAlias(String alias) {
...
+                    item.setEntrezGeneId(row.getEntrezGeneId());
+                    item.setHugoSymbol(row.getHugoSymbol());
+                    return item;
+                }
+            );
+
+            if ("GRCh37".equalsIgnoreCase(row.getReferenceGenome())) {
+                setIsoformWithWarning(vm, row.getEnsemblTranscriptId(), true);
+            } else if ("GRCh38".equalsIgnoreCase(row.getReferenceGenome())) {
+                setIsoformWithWarning(vm, row.getEnsemblTranscriptId(), false);
```
calvinlu3, line 175: Can use ReferenceGenome enum

calvinlu3 (approved):

jfkonecn:
I'm going to use the database directly so we can avoid changing the version on the transcript service

